PR #4, opened 2015-12-13
Add support for OS X Yosemite and Cypress Gen3 I2C trackpad.

coolstar:
Adds support for OS X Yosemite and preliminary support for the Cypress Gen3 I2C Trackpad on Haswell chromebooks.


alexandred:
Thanks!

PR #5, opened 2015-12-14
Fix up issues with Cypress driver on OS X.

coolstar:
I've fixed up issues with the Cypress driver. I also created a temporary VoodooHIDMouseWrapper to avoid breaking VoodooMouseWrapper for the HID touch screen devices. Trackpad is fully functional in OS X as a mouse. Will need more experimentation to get working as a trackpad and not mouse but it works.


alexandred:
Thanks!

PR #8, opened 2016-03-30
Update Cyapa Gen3 trackpad driver.

coolstar:
Updated the Cyapa Gen3 trackpad driver with the latest code from crostrackpad 3 beta 8.

Still has an issue when rebooting between windows and OS X but works fine on a cold boot of OS X.


alexandred:
Thanks!

PR #9, opened 2016-03-31
Fixes for Cypress Gen3 driver stability.

coolstar:
Fix issue with unloading kext with Cyapa driver loaded and fix issue when rebooting from Windows -> OS X


alexandred:
Thanks!

PR #11, opened 2016-04-09
Fix power management issue with Cyapa driver. Add initial Atmel Touch Screen Driver.

coolstar:
Update Cypress Trackpad Driver to fix power management issue. Add initial port of Atmel Touch Screen code from crostouchscreen2.3. Touch Screen driver causes freeze for some reason, so disabled for now.

Also check the Hardware ID's for CYAP0000 and ATML0000 as those are only on the chromebooks and need the special drivers.


alexandred:
Thanks!

PR #12, opened 2016-04-15
Add sleep/resume support to I2C bus driver and Cypress driver.

coolstar:
Add sleep/resume support to I2C bus driver and Cypress driver. Sleep/wake is working now. Tested with Cypress pad in Acer C720 put to sleep and woken up.

C720 still can't be woken up using the trackpad, but the trackpad works after a sleep/wake cycle once the OS is resumed.


alexandred:
Thanks!: TODO: PM for i2c-hid

PR #18, opened 2016-06-30
Refactor CSGesture code to common file & use common mouse wrapper for CSGesture.

coolstar:
Refactors CSGesture code to common file and use a common mouse wrapper for all CSGesture trackpads to reduce code duplication. Use common csgesture file for Cypress/Elan Trackpads.


alexandred:
Thanks!

PR #19, opened 2016-09-07
Use IOHIPointing for CSGesture-based trackpads, add inertia scroll. Update Elan code.

coolstar:
This patch refactors code for all CSGesture based trackpads (currently Cypress and Elan).

Also adds smooth scrolling and inertia scroll (ported from crostrackpadscroll on Windows) to both Cypress and Elan trackpads so scrolling is closer to real Macbooks' scrolling.

Elan trackpads' code has been updated to the latest from crostrackpad3-elan to resolve resolution issues on devices other than the Acer C720 and C740 chromebooks.

This also adds initial support for the trackpad settings pane (currently horizontal scroll, tap to click and tap drag can be toggled via Apple's settings pane, with more to come). The trackpad settings pane is the same one that shows up when VoodooPS2Controller is used on a Synaptics PS2 trackpad.

Finally I cleaned up compiler warnings with CSGesture, Cypress & Elan Trackpads and the Atmel touch screen drivers.


alexandred:
Thanks!

PR #20, opened 2016-09-14
Code cleanup; Skylake PCI support; ACPI fallback; more code comments; better error logging; ELAN1000 multitouch

coolstar:
I have made several cleanups to the I2C bus code on platforms that use ACPI (Haswell/Broadwell) so that we only need to keep 1 reference to the IOService provider around, casting in the few places we need IOACPIPlatform for the ACPI calls. The ACPI config calls now properly return false if there's a problem querying ACPI, and a fallback method is used if the ACPI calls fail to set hardcoded known values for haswell/broadwell.

For Skylake support, I have cleaned up and imported code from maximevince's repo, making minimal changes to VoodooI2C.cpp (sharing as much common codebase as we can between ACPI and PCI, and for the minimal PCI specific code, it is in the separate VoodooI2CPCI.cpp) file. As for the Skylake hack, I have added an enum that describes whether the i2c bus is running in ACPI or PCI mode and only applies the skylake hack if it's in PCI mode. Some adjustments to Info.plist were needed to add the Skylake PCI ID's and to allow the kext to load with both PCI and ACPI configurations.

I have then brought over some other improvements from Linux. Aside from adding additional code comments, I have added better error logging regarding i2c transaction aborts. (It now detects the reason for the abort and prints out a description based on the structs in the Linux driver).

We have tested these updates on the Haswell Acer C720 Chromebook (Cypress trackpad + Atmel touch screen), Broadwell Acer C740 Chromebook (Elan Trackpad), blankmac's Skylake HP Elite X2 1012 (Wacom touch + stylus), and alexviean's Skylake Asus K501UB-DM039D (Elan Trackpad) and it works on all 4 configurations 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ The following devices are confirmed to work with the current iteration of Voodoo
   * This uses the proprietary interface `VoodooI2CAtmelMxtScreenDevice`.
 * ELAN0000 trackpad (Broadwell Chromebooks)
  * This uses the proprietary interface `VoodooElanTouchpadDevice`
+* ELAN1000 trackpad (Skylake Asus K501UB-DM039D)
+ * This uses the proprietary interface `VoodooElanTouchpadDevice`
 
 # Helping out and testing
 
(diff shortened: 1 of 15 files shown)


coolstar:
As of the last commit, we now have full multitouch working on ELAN1000 (Asus K501UB-DM039D)

Nice to finally have multitouch on a non-chromebook i2c trackpad :)

alexandred:
Thanks!

PR #21, opened 2016-10-03
Fix Atmel Touch Screen input in OS X and force Elan back into absolute mode on wake

coolstar:
1. Remove HID Confidence and In Range bits from Atmel driver since it's not needed
2. Sanitize Atmel input so the Atmel driver works and inputs touch points correctly
3. Fix "Sticky" touches on Atmel -- fully working touch screen in OS X (finally) :smile: 
4. Force Elan touchpad into absolute mode on wake from sleep to fix suspend issue on ELAN1000.


alexandred:
Thanks!

PR #22, opened 2016-10-12
Minor code cleanups. Allow use on OS X Mavericks.

coolstar:
Do some minor code cleanup to try to follow OS X kext coding conventions.

Allow use of the kext on OS X Mavericks; installed Mavericks on a partition here to run some tests and I found the kext works fine on Mavericks with some small minor changes.


alexandred:
Thanks!

PR #23, opened 2016-10-27
Added Asus X556UF to supported devices

sahaRatul:


alexandred:
Please add this to the Compatibility page of the wiki instead.

PR #24, opened 2016-11-25
Re-add some code from Max's branch

MacForceOne:
Changes how the hid device is initialized, which allows it to properly initialize ELAN1200 trackpads.
Also adds input filtering which is required by some devices, ASUS laptops in particular


coolstar:
filtering repeated reports like this would break any digitizers that use absolute coordinates like touch screens. This is specific to ELAN1200

timofonic:
@coolstar @MacForceOne @alexandred @RehabMan Has this been fixed in current version? I see there are many VoodooI2C forks, I'm confused about what one to use and such :cry:

alexandred:
@timofonic This has been fixed in the rewrite. We don't currently have an ETA for its release unfortunately.

PR #25, opened 2016-12-10
Sierra pref pane enable

blankmac:
Checks the Darwin version before applying the properties to enable the preference pane.  Adds "precise" scrolling properties (tested and working)  The preference pane is enabled, however, the only tester reports that he has to toggle the click setting on reboot.  On my machine, the click setting is persistent, but unused so I can't test.


alexandred:
Thanks!

PR #27, opened 2017-03-04
Force enabling 'Tap to click' and 'dragging'

ben9923:
Until read from preferences would work, it's better to have it enabled.
Changes from System Preferences won't take effect now.


coolstar:
There's probably a better solution for Sierra, but this should do for now.

sooxiaotong:
Pls fix it , thanks :3

ben9923:
Fix was found by @kprinssu 
PR #32 
No need to force-enable it anymore.

PR #30, opened 2017-03-29
Properly initialise the RMI device (as per Linux kernel)

kprinssu:
I noticed that the Linux sets a different RMI mode (see https://github.com/torvalds/linux/blob/master/drivers/hid/hid-rmi.c#L480) and this seems to get basic multitouch working on my laptop.

You should also note that for proper device support, you will need to modify the if/else statement in VoodooI2C.cpp

### DIFF
--- a/VoodooI2C/VoodooSynapticsRMITouchpadDevice.cpp
+++ b/VoodooI2C/VoodooSynapticsRMITouchpadDevice.cpp
@@ -936,7 +936,7 @@ int VoodooSynapticsRMITouchpadDevice::rmi_set_mode(uint8_t mode) {
 int VoodooSynapticsRMITouchpadDevice::rmi_populate() {
     int ret;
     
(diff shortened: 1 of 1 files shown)


kprinssu:
Right now the best way to enable multitouch for Synaptics RMI devices is to use a DSDT edit.

@blankmac suggests:
into device label TPD0 insert begin Name (_HID, "SYNA0000")\n end;

OR

into device label TPL1 insert begin Name (_HID, "SYNA0000")\n end;


Personally, I had to edit the TPL1 device's _INI method.

coolstar:
@kprinssu SYNA0000 is the device ID for Dell Chromebook 13. I'd suggest sending a pull request to add your device ID here: https://github.com/alexandred/VoodooI2C/blob/master/VoodooI2C/VoodooI2C.cpp#L727

PR #32, opened 2017-04-08
Fix preference pane issue on macOS Sierra

kprinssu:
Rehabman uses OSBoolean instead of OSNumber (see https://github.com/RehabMan/OS-X-Voodoo-PS2-Controller/blob/master/VoodooPS2Trackpad/VoodooPS2SynapticsTouchPad.cpp#L2518)

Not tested on macOS versions lower and including 10.11 (El Captain), but should work nonetheless.


coolstar:
Looks good to me.

PR #34, opened 2017-04-20
Fix tap/clicks interfering with momentum scroll

kprinssu:
2F press and then a scroll gesture causes an odd bug where context menu appears repeatedly.


kprinssu:
Currently this inherently flawed and I will be reworking this PR.

PR #36, opened 2017-04-20
allow _DSM or XDSM for HID data acquisition; and avoid KP if both are missing

RehabMan:
It is very common to rename _DSM methods in DSDT (and other ACPI files) to XDSM, so that additional _DSM methods can be added (either through patching or via add-on SSDTs) without conflict.

I use hotpatch for _DSM->XDSM (via config.plist/ACPI/DSDT/Patches).  Doing _DSM->XDSM rename, excluding the I2C from the rename would be difficult (to say the least).

This poses a problem with this kext, as it KPs when the expected _DSM method is not found.  Ideally, a kext should not KP due to user error.

I made two improvements here:
- no KP if the expected _DSM does not evaluate without error
- both _DSM and XDSM are allowed

This allows _DSM or XDSM (compatible with _DSM->XDSM hotpatch), and should the user mistakenly remove all _DSM methods from their DSDT, they will not get a KP...

I have no I2C hardware, but I proposed this change for other users...

### DIFF
--- a/VoodooI2C/VoodooI2CHIDDevice.cpp
+++ b/VoodooI2C/VoodooI2CHIDDevice.cpp
@@ -245,12 +245,17 @@ int VoodooI2CHIDDevice::i2c_hid_acpi_pdata(i2c_hid *ihid) {
     params[2] = OSNumber::withNumber(0x1, 8);
     
     ihid->client->provider->evaluateObject("_DSM", &result, params, 3);
+    if (!result)
+        ihid->client->provider->evaluateObject("XDSM", &result, params, 3);
+
 
     OSNumber* number = OSDynamicCast(OSNumber, result);
-    
-    ihid->pdata.hid_descriptor_address = number->unsigned32BitValue();
-    
-    number->release();
+    if (number)
+        ihid->pdata.hid_descriptor_address = number->unsigned32BitValue();
+
+    if (result)
+        result->release();
+
     params[0]->release();
     params[1]->release();
     params[2]->release();


coolstar:
Shouldn't be any problem with handling XDSM as well as _DSM. However, the way the I2C HID spec works, the HID driver should report an error and exit if it can't find the offset for the I2C HID descriptor. Trying to continue with enumeration without the HID descriptor offset would result in undefined behavior (either a kernel panic later in the kext, or worse leaving the i2c device in a state where it needs a hard shutdown to recover)

I would have to test and see what the generic HID driver does myself sometime (I use the Elan i2c driver myself here)

RehabMan:
I didn't look at upstream code on the call stack (code calling VoodooI2CHIDDevice::i2c_hid_acpi_pdata).
I was just trying to avoid the very obvious bad code causing panic.

If there is an issue where calling code has a problem with ihid->pdata.hid_descriptor_address not being initialized, that code should be changed too.
Graceful exit (eg. failure to start) with appropriate logs preferred over KP.

alexandred:
Hi RehabMan,

Many thanks for the PR. Yes I agree that the codebase is a mess and various elements desperately need refactoring, I unfortunately no longer have the time these days to do so myself.

I will leave it to @coolstar to merge this when he is happy with it.

blankmac:
Tested with my Wacom fork (which uses the generic HID code) and the result is KP if neither _DSM or XDSM are present.

ben9923:
Tested with a user's ELAN1200 (MacForceOne's fork) that has XDSM - working well.

alexandred:
I am going to accept this PR now, will leave a mental note regarding @blankmac's crash. Thanks!

PR #37, opened 2017-04-27
Fix 2F pref and scroll bug

kprinssu:
This will fix the scroll from activating when 2F physical button are pressed.

Tested on my laptop and works properly.


kprinssu:
Just a fixed regression in the system preference pane support (also removed some chatty debug print statements). Looks like its best that both casts are left in place.

kprinssu:
Going to be re-working this PR.

alexandred:
Thanks!

PR #40, opened 2017-07-05
Default tapToClickEnabled to true

ben9923:
For tap to click inside the lock screen, until settings are read from user's preferences (after login).


alexandred:
Thanks!

PR #42, opened 2017-07-17
Fix verbose boot issues

ben9923:
Commenting out 2 logging lines which caused kernel panics in a few devices when booting verbose.


alexandred:
Thanks!

PR #92, opened 2018-09-21
Allow waiting for the GPIO controller

ben9923:
Refactored VoodooI2CDeviceNub::getGPIOController() so it waits (up to 1 second) for the GPIO controller if not available.


alexandred:
Thanks!

PR #106, opened 2018-09-27
drag window

dedicahyadi:
thanks alex for the quick update 2.1.3 it's works, but there is still a little bit problem with drag window, it seems to be caused by touching the click area. can you disable touch in the click area. the point is that it can't drag the window when it touches the click area with a finger. sorry bad english

thank in advance


alexandred:
Yes this is due to scaling which is an ongoing issue which we do not yet have a solution for.

dedicahyadi:
> Yes this is due to scaling which is an ongoing issue which we do not yet have a solution for.

okay, thank you sir.

alexandred:
I just realised you opened this as a PR, it should be an issue instead.

Issue #132, opened 2018-11-05
Mouse cursor randomly jumps to top corner screen


cookievu:
My trackpad is Elan 1200, use VoodooI2C v2.1.4 & VoodooI2CHID on Mojave.
Everything works well, except for the cursor that usually jumps to the top corner of the screen.
Is there a way to solve this problem?

Jasper0819X:
same problem here with elan 1200 and mojave 10.14.1

alexandred:
Provide your troubleshooting archives as requested on the troubleshooting page: https://voodooi2c.github.io/#Troubleshooting/Troubleshooting.

cookievu:
This is my archive. Hope can find the cause.
https://drive.google.com/open?id=1gg5J52WiVGS0bRz1YS-D8mWRbpCgINy9

Jasper0819X:
@cookievu Do you have VoodooI2CELAN in your Clover's kexts folder?

coolstar:
Try commit ecabd7461c549f29a8a5b776a8b10041232dff25, if it's a scaling issue, should be fixed now.

andrewwyatt:
I built it before work today, and just now had some time to test.  There's a noticeable improvement in cursor speed and multi-touch input but unfortunately the cursor jumping to the right corner problem still persists for me.

hohoho123456:
i have the same problem with ELAN trackpad. other than that and 4-5 finger gesture support, it's perfectly working.

LeeBinder:
@coolstar : I'm having [two odd issues with 2.1.4]([url](https://github.com/alexandred/VoodooI2C/issues/108#issuecomment-451006810)), with both, current official alexandre release and also self compiled in 10.13.6 (even with 10.13.6 as target OS in the workspace Xcode serttings).

v. 2.0.3 is close to perfect on my Asus Vivobook with ELAN 1300 - performs even smoother than the Windows driver! The ONLY thing which annoys me in 2.0.3 is this random but frequent pointer jumping.

After reading what you wrote, I'm now wondering: is it possible for you to also optimize the scaling in the corresponding section of the 2.0.3 code?

Looking at all the issues ever since the engine change, it seems a worthwhile idea to not completely abandon 2.0.3, but rather to also implement those 2.1.4 fixes which are possible and helpful.

It would make my day if you could see if you can also fix the scaling in the 2.0.3 source code. I'd be more than happy to compile myself and report back :)

LeeBinder:
[...]

PR #169, opened 2019-04-06
Apply patch from issue #132

andrewwyatt:
This is a PR to merge xiqin's patch from issue #132.  It has soaked for a while now across a few different systems and seems to completely correct the jumping cursor condition for those of us who have been using it.


alexandred:
Sounds good to me, thanks!

Issue #148, opened 2019-01-15
When the display is rotated 90 degrees, the touchscreen gestures do not rotate


ghost:
On a touch screen convertible, when you rotate the display using System Preferences -> Command + Option -> Displays -> Rotation -> 90 degrees, touch screen single finger tracking follows the display rotation properly but gestures do not change orientation and continue to operate following the original orientation.

jslegendre:
This is a bug from macOS itself. For whatever reason Apple has decided(forgotten?) to support rotated screens for a while now.  Real Apple laptops haven't been able to easily rotate 90 degrees since the retina screens came out (late 2012) and the bug is still present even on the latest version of Mojave.  

Apple could fix this quite easily but it would be a very involved and painstaking process for a third-party.

alexandred:
@jslegendre Thank you for confirming this is a problem on Apple's side (I always suspected it was). It is actually not too difficult for us to fix by checking the rotation ourselves and manually swapping x and y coordinates before we send to the multitouch stack. We used to do this with CSGesture (and it worked a treat) but I haven't had the time to port it to the native engine as of yet.

blankmac:
@andrewwyatt Please test and comment - https://github.com/alexandred/VoodooI2C/pull/172
Thanks.

andrewwyatt:
@blankmac, I can confirm that your patch corrects the issue on my ASUS Q405UA 2-in-1.

blankmac:
Great, thank you.

PR #172, opened 2019-04-09
Fix multitouch input when screen is rotated.

blankmac:
Tested and working on my HP Elite X2 @ 0,90,180,270 degrees of rotation.


blankmac:
Confirmed to work by issue opener - https://github.com/alexandred/VoodooI2C/issues/148

Issue #187, opened 2019-05-08
VoodooI2CHID leads to a kernel panic with OpenCorePkg


anonymous-writer:
Hello developers,
With Clover the Kext still works as a dream. Thanks to your very good work.

Now I have tried to use the Kext under OC.
https://github.com/acidanthera/OpenCorePkg

I know that's all still in development but I would like to post the result, since the I2C Kext is the only one which leads to a kernel panic with OC. The kernel panic always occurs only by reboot oder shutdown of the laptop.
Maybe someone would like to look at the bug report and have any idea why this kernel panic happens.

Here are some more information:

GPIO is not possible with my Zenbook UX330UAK. Asus BIOS is buggy und CPU frequency is going up du max level with GPIO.
Therefore, for my ASUS only polling is possible.

VoodooI2C + VoodooI2CHID not injected > Laptop works perfect.

Only VoodooI2C is installed > Laptop works perfect.

VoodooI2C + VoodooI2CHID are installed to /L/E > Not possible with OC since the SIP is fully activated. Must be deactivated by boot in the recovery mode and I did not want to do that .

Kernelpanik also happen with the official version from mojave 10.14.4.

[debug_30391.zip](https://github.com/alexandred/VoodooI2C/files/3156107/debug_30391.zip)

stevezhengshiqi:
Exactly the same problem here.

vit9696:
Hello,

Fritz asked me to do a brief check on this, and here is the details I can provide. The kernel panic happens due to a use-after-free in IOHIDEventService in [IOHIDFamily](https://opensource.apple.com/tarballs/IOHIDFamily):

```C++
void IOHIDEventService::close(IOService *forClient, IOOptionBits options)
{
    _commandGate->runAction(OSMemberFunctionCast(IOCommandGate::Action, this, &IOHIDEventService::closeGated), forClient, &options);
}
```

`_commandGate` is a macro pointing to `_reserved→commandGate`, and given the kernel panic `_reserved` was actually freed. The only method that does this is `IOHIDEventService::free()`.

I am not sure how exactly this can happen, but it seems to be a race condition during driver unloading in either IOHIDFamily itself, or perhaps VoodooI2CHID. While it is strange that it reproduces reliably with our kext injector, we checked the code on our side, and OpenCore itself cannot be any relevant: the bug must have existed long before, and it just became more visible.

[...]

PR #195, opened 2019-06-09
Fix resource leaks in and refactoring in various VoodooI2C classes

kprinssu:
This should help address https://github.com/alexandred/VoodooI2C/issues/187 but does not fix it.

THe changes are from @acidanthera's VoodooPS2 port.


Multitouch Support/Native/VoodooI2CMT2SimulatorDevice.cpp
```diff
@@ -437,9 +437,10 @@ IOReturn VoodooI2CMT2SimulatorDevice::getReport(IOMemoryDescriptor* report, IOHI
     Boolean getReportedAllocated = true;
     
     OSData* get_buffer = OSData::withCapacity(1);
-    
-   if (!get_buffer || (report_id == 0x1 && !new_get_report_buffer)) {
-        return kIOReturnNoResources;
+
+    if (!get_buffer || (report_id == 0x1 && !new_get_report_buffer)) {
+       OSSafeReleaseNULL(get_buffer);
```
kprinssu: No need to free the buffer here. It's going to short circuit and the release will never get called since get_buffer will always be NULL.

kprinssu:
At this point, it looks like this PR is turning into a big clean up of the resource handling of VoodooI2C. 

There's a still few issues with VoodooI2C when unloading. I am not sure if it's a leak in the satellite kext (VoodooI2CHID) or in the controller classes.

alexandred:
Does this PR help address CPU usage issues in kernel_task when using VoodooI2C? (One day I traced the CPU usage problem to somewhere in the controller drivers so it can't be an issue in any of the satellites).

kprinssu:
I am not sure if this PR fixes that bug. I’ll take a look at it on it after fixing a few remaining leaks that I found.

ben9923:
Merged!
Great work on that @kprinssu :)

PR #224, opened 2019-10-26
Correct IOCommandGate::commandSleep call

al3xtjames:
`IOCommandGate::commandSleep` has the following signatures:

```cpp
virtual IOReturn commandSleep(
 void *event,
 AbsoluteTime deadline,
 UInt32 interruptible);

virtual IOReturn commandSleep(
 void *event,
 UInt32 interruptible);
```

Previously, `commandSleep` was called with 2 arguments, with the deadline
being passed in as `interruptible`, which is incorrect. Explicitly specify
`THREAD_UNINT` as `interruptible` to use the first signature and correctly
specify the deadline.


ben9923:
Thanks!

al3xtjames:
No problem :)

Issue #236, opened 2020-01-02
Comet Lake Unknown Synopsys component type: 0xffffffff 


thefiredragon:
Is there a way to fix Unknown Synopsys component type: 0xffffffff on Comet Lake  i7 1051u?
I could provide logs. 
Think it makes no sense to try patch my DSDT before this is solved or I am wrong here at this point?

best Regards

ben9923:
Hey there,
Unfortunately I'm not aware of anyone with 10th Gen hardware who could make VoodooI2C properly load :(
This could be some sort of an ACPI issue that patching (no idea what exactly) might solve.

thefiredragon:
Could we debug this?

fewtarius:
Several of us have 10th gen hardware and would love to help debug this.  The problem appears to begin within VoodooGPIO.  In the data I provided, you can see it is failing on an interrupt request attempting to instantiate VoodooGPIOCoffeeLakeLP on a comet lake machine.

https://github.com/alexandred/VoodooI2C/issues/234

thefiredragon:
Okay thanks, how I can create a debug report?
I don't have a kernel panic when I use voodooi2c perhaps this could be helpful,
best Regards

fewtarius:
I'd expect that there will be multiple problems around supporting comet lake, I'm chiming in with a me-to.  You can build a debug package by following this:

https://github.com/alexandred/VoodooI2C/blob/master/Documentation/Troubleshooting.md

3ig:
So ... still no solutions for this problem. Does anyone have any progress ?

neblogas:
I have the same CPU i7 10510U
Lenovo S540 14
Installer does not boot with VoodooI2C.kext
Without the kext I can boot into installer but the trackpad does not work.
In README I can see that these ID's are fully supported:
Hardware ID:                                                                    PCI\VEN_8086&DEV_**02E8**&SUBSYS_380417AA&REV_00
Hardware ID:                                                                    PCI\VEN_8086&DEV_**02E9**&SUBSYS_380317AA&REV_00

However I have also this one:
Hardware ID:                                                                    PCI\VEN_8086&DEV_**02EA**&SUBSYS_380617AA&REV_00

which is not in the supported list.
Maybe thats the issue? Is there a way to support it?

williambj1:
@neblogas 

[...]

PR #245, opened 2020-02-18
Fix Unknown Synopsys component type issue in Comet Lake

startpenghubingzhou:
This commit has fixed the issue https://github.com/alexandred/VoodooI2C/issues/236 . This work is based on https://www.notion.so/Using-VoodooI2C-on-comet-lake-cpu-e-g-i5-10210u-142930887087445eaa533120455da5dc , which will force D0 in Comet Lake. And I completed it to check the platform to avoid this patch to be applied in other platforms. See code comments for more details.

### DIFF
--- a/VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CPCIController.cpp
+++ b/VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CPCIController.cpp
@@ -12,11 +12,43 @@
 OSDefineMetaClassAndStructors(VoodooI2CPCIController, VoodooI2CController);
 
 void VoodooI2CPCIController::configurePCI() {
+    char tmp[2];
+    const char kCometLakeflag[2] = {'2', 'e'};
+    
     IOLog("%s::%s Set PCI power state D0\n", getName(), physical_device.name);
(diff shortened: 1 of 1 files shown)


kprinssu:
There must be a better way of determining if the chipset is indeed Cometlake rather comparing the device name. Perhaps implementing a querying the hardware or a boolean in a DSDT/SSDT patch?

startpenghubingzhou:
> There must be a better way of determining if the chipset is indeed Cometlake rather comparing the device name. Perhaps implementing a querying the hardware or a boolean in a DSDT/SSDT patch?

this  commit has conflict with master so I  ll close it and send a new pr again.

startpenghubingzhou:
@kprinssu  https://github.com/alexandred/VoodooI2C/pull/246 I ll reply here

phu54321:
Fix author here. My notion document is tied to the school account I've graduated from. It'd be safer to write a backup here.

- With ubuntu `sudo lspai -vv` shows where the power management state is stored in PCI config space. In case of CFL its 0x80.
- enablePowerState wont wake touchpad to D0. Probably macOS bug or some unknown internal things.
- By overwriting the PCI config space itself we can force touchpad to wake. (D0 state) This fixes the bug.

PR #257, opened 2020-03-22
Automatically build VoodooInput

hieplpvip:
This pull request automatically builds VoodooInput and copies it to `Plugins` folder inside `VoodooI2C.kext`, just like `VoodooGPIO`

### DIFF
--- a/VoodooI2C/VoodooI2C.xcodeproj/project.pbxproj
+++ b/VoodooI2C/VoodooI2C.xcodeproj/project.pbxproj
@@ -538,7 +538,7 @@
 				AC2603AC1F2F294000CF238F /* Frameworks */,
 				AC2603AD1F2F294000CF238F /* Headers */,
 				AC2603AE1F2F294000CF238F /* Resources */,
-				AC0955921F4EE9B00052E343 /* Add GPIO and Services Plugin */,
+				AC0955921F4EE9B00052E343 /* Add GPIO, Services and VoodooInput Plugin */,
 				ACA4A00D21E25F0E00A91B61 /* Linting */,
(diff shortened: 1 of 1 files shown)


alexandred:
@vit9696 I agree that VoodooI2C and VoodooPS2 should be regarded as plugins of VoodooInput. However, the way VoodooI2C is built means that VoodooInput should also be regarded as a plugin of VoodooI2C.

By default all our releases contain a freshly built copy of all plugins for ease of use to the end user (and also for ease of release). I recognise, however, that in order to ensure homogeneity across the userbases for VoodooI2C and VoodooPS2 there should be a singular release source for VoodooInput.

As such I am happy to forgo automatic builds for VoodooInput and to setup our continuous integration to grab the latest release of VoodooInput from AcidAnthera's repository and bundle that with our releases. This way we can maintain our current release structure while respecting the shared use of VoodooInput. Does this sound okay to you?

vit9696:
Yes, this sounds fine to me. As long as we use the same binaries this should work reliably. Maybe mention VoodooInout version used in the release notes.

vit9696:
Actually, you may build something like the script we have for Lilu for plugin compilation:
https://github.com/acidanthera/Lilu/blob/master/Lilu/Scripts/bootstrap.sh
https://github.com/acidanthera/WhateverGreen/blob/master/.travis.yml

kprinssu:
I am against us building VoodooInput, the point of VoodooInput was to abstract away the MT2 dependency. By introducing this PR, we will no be explicitly re-introducing the dependency back into the project. We should either implement a `wget` or `curl` on the latest release in the CI flows or we something similar.

This way, we can maintain the latest version of VoodooInput support.

PR #266, opened 2020-04-01
Optimise Polling Mode

lvs1974:
Notify VoodooI2CHIDDevice when touch pad data were handled and sent to VoodooInput.
In VoodooI2CHIDDevice it will be used to adjust polling mode timer interval, it will be dependent on touch pad usage.
Change for VoodooI2CHIDDevice will be committed in a separate pull request.


ben9923:
Hey,
Please note VoodooI2C changes should be compatible with all satellites, not just VoodooI2CHID.
Core kext should be generic...

lvs1974:
> Hey,
> Please note VoodooI2C changes should be compatible with all satellites, not just VoodooI2CHID.
> Core kext should be generic...

Yes, you are right!
Actually it can be implemented only in VoodooI2CHID!
So, this PR will be closed, only one PR https://github.com/alexandred/VoodooI2CHID/pull/26 can be merged to get optimised polling mode.

PR #268, opened 2020-04-02
Update VoodooI2CHID

kprinssu:
Contains polling mode improvements from @lvs1974


marianopela:
It's kind of laggy now. I can notice it when moving it slowly

lvs1974:
> It's kind of laggy now. I can notice it when moving it slowly

Could you put your boot log here taken with the following command in terminal:
log show --last boot --predicate 'process == "kernel"' | grep -i intel > ~/Desktop/log.txt

And video would be more informative.

kprinssu:
@marianopela please move the discussion to a Github issue rather posting here.

PR #272, opened 2020-04-05
CI updates

kprinssu:
- VoodooInput now builds with VoodooI2C (VoodooInput maintainers are moving towards the developers building and distributing their own binaries, see https://github.com/acidanthera/VoodooInput/commit/7e341d9a7261a5efacff1368e4f4ca748ab14773)
- Update the minimum supported version of macOS to macOS 10.13 (macOS 10.12 reached end-of-life in October 2019)


vit9696:
First is wrong. You should never distribute VoodooInput binaries *built* by VoodooI2C.

Second is undesired. Many machines got sleep broken in 10.13, so they still stay on 10.12. I do not think end-of-life should be used as a reason to drop older macOS versions. For example, we target 10.8 in Acidanthera for most drivers and 10.11 for VoodooPS2.

alexandred:
@kprinssu As agreed with @vit9696 in https://github.com/alexandred/VoodooI2C/pull/257, we should avoid building VoodooInput with our CI and instead grab the latest binary from AcidAnthera's repository and bundle it with our releases.

kprinssu:
@alexandred @vit9696 duly noted. I am going to revert the changes to min. build version and the changes to VoodooI2C building the kext.

PR #287, opened 2020-04-15
Add an option to disable Force Touch emulation

Goshin:
- Add an option in info.plist to disable Force Touch emulation to solve drag bug on some devices.

It is set enabled in default since I think in most cases it can be disabled in system preference painlessly.

@lvs1974 @tmbt78 plz test it. The property path is `info.plist / IOKitPersonalities / Native Multitouch Engine / EnableForceTouchEmulation`.

### DIFF
--- a/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
+++ b/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
@@ -88,22 +88,10 @@ bool VoodooI2CNativeEngine::start(IOService* provider) {
     setProperty(kIOFBTransformKey, 0ull, 32);
     setProperty("VoodooInputSupported", kOSBooleanTrue);
 
-
-    OSDictionary* nameMatch = IOService::serviceMatching("AppleMultitouchDevice");
-    appleMultitouchDevice = waitForMatchingService(nameMatch, 1e9);
-    if (!appleMultitouchDevice) {
-        IOLog("%s::fail to get AppleMultitouchDevice\n", getName());
-    }
-    nameMatch->release();
-
     return true;
 }
 
 bool VoodooI2CNativeEngine::isForceClickEnabled() {
-    if (!appleMultitouchDevice) {
-        return lastIsForceClickEnabled;
-    }
-
     AbsoluteTime now_abs;
     uint64_t diff_ns;
     clock_get_uptime(&now_abs);
@@ -115,7 +103,7 @@ bool VoodooI2CNativeEngine::isForceClickEnabled() {
     }
 
     // Blocking reading here takes about 10 microseconds
-    OSDictionary* dict = OSDynamicCast(OSDictionary, appleMultitouchDevice->getProperty("MultitouchPreferences", gIOServicePlane, kIORegistryIterateRecursively));
+    OSDictionary* dict = OSDynamicCast(OSDictionary, this->getProperty("MultitouchPreferences", gIOServicePlane, kIORegistryIterateRecursively));
     if (!dict) {
         return lastIsForceClickEnabled;
     }
--- a/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
+++ b/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
@@ -54,7 +54,7 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
 
         // Force Touch emulation
         // The button state is saved in the first transducer
-        if (((VoodooI2CDigitiserTransducer*) event.transducers->getObject(0))->physical_button.value()) {
+        if (isForceTouchEmulationEnabled && ((VoodooI2CDigitiserTransducer*) event.transducers->getObject(0))->physical_button.value()) {
             inputTransducer->supportsPressure = true;
             inputTransducer->isPhysicalButtonDown = 0x0;
             inputTransducer->currentCoordinates.pressure = 0xff;
@@ -87,6 +87,12 @@ bool VoodooI2CNativeEngine::start(IOService* provider) {
     
     setProperty(kIOFBTransformKey, 0ull, 32);
     setProperty("VoodooInputSupported", kOSBooleanTrue);
+
+    isForceTouchEmulationEnabled = true;
+    OSBoolean* enableForceTouchEmulationProperty = OSDynamicCast(OSBoolean, this->getProperty("EnableForceTouchEmulation"));
+    if (enableForceTouchEmulationProperty) {
+        isForceTouchEmulationEnabled = enableForceTouchEmulationProperty->getValue();
+    }
 
     return true;
 }
--- a/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
+++ b/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
@@ -54,7 +54,9 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
 
         // Force Touch emulation
         // The button state is saved in the first transducer
-        if (isForceTouchEmulationEnabled && ((VoodooI2CDigitiserTransducer*) event.transducers->getObject(0))->physical_button.value()) {
+        if (isForceTouchEmulationEnabled
+            && ((VoodooI2CMultitouchInterface*) getProvider())->isForceClickEnabled
+            && ((VoodooI2CDigitiserTransducer*) event.transducers->getObject(0))->physical_button.value()) {
             inputTransducer->supportsPressure = true;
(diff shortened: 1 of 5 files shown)


lvs1974:
@Goshin, I have tested it:
When option EnableForceTouchEmulation in Info.plist is set to True, I observe these issue with dragging and text selection.
When option EnableForceTouchEmulation is set to False, everything is fine!
Thank you!

kprinssu:
Is there any way you can programmatically do this rather have this as a plist configuration?

Goshin:
Just the same as VoodooPS2. Do you mean using a boot arg?

Goshin:
We can get the Force Touch property in `setProperties()`, which is a better way to get the option. I am still testing.

kprinssu:
@Goshin `setProperties` is quite vague, where exactly do you have in mind?

Goshin:
> @Goshin `setProperties` is quite vague, where exactly do you have in mind?

I meant we can obtain the properties regarding the Trackpad preference in 

```
IOReturn VoodooI2CMultitouchHIDEventDriver::setProperties(OSObject * properties) {
```

The properties `ActuateDetents` and `ForceSuppressed` are related to the `Force Click and haptic feedback` option in `Preferences -> Trackpad`.
`ActuateDetents` is true and `ForceSuppressed` is false when the option is enabled, but `ForceSuppressed` is not available when booting. 

We can get the property and pass it to the multitouch interface, https://github.com/Goshin/VoodooI2CHID/commit/1a23d4ddf991052c08c019ef9bee6f3126106c14
so the Force Touch emulation can be enabled or disabled accordingly.

hieplpvip:
@Goshin That would be satellite-specific. Are there any ways to get those properties directly from `VoodooI2CMultitouchInterface`?

Goshin:
I agree with you it is a little bit ugly, but when I test, the properties can only be obtained in `VoodooI2CMultitouchHIDEventDriver::setProperties`. Seems those HIDEventServiceProperties properties are only set on IOHIDInterface. Any idea about it?

Or we can pass the property to `VoodooI2CMultitouchInterface`?

```cpp
IOReturn VoodooI2CMultitouchHIDEventDriver::setProperties(OSObject * properties) {

    ...    

    if (multitouch_interface) {
        multitouch_interface->setProperties(properties);
    }

    return super::setProperties(properties);
}
```

Goshin:
The new commit process the properties in the multitouch engine, which is a more intuitive procedure.

The HIDDevice will pass the properties to the multitouch interface, so the registered engines can obtain the properties they are interested in. https://github.com/alexandred/VoodooI2CHID/compare/master...Goshin:ft-prop

A question, should we keep the configuration property in plist?

hieplpvip:
`A question, should we keep the configuration property in plist?`
Using `System Preferences` is easier than changing the settings in plist, and we don't have to do it every time a new version is available. I think you should remove that property.

Mateo1234454545:
> `A question, should we keep the configuration property in plist?`
> Using `System Preferences` is easier than changing the settings in plist, and we don't have to do it every time a new version is available. I think you should remove that property.

Unfortunately, changing just the option in settings doesn't have any outcome.

Goshin:
> > `A question, should we keep the configuration property in plist?`
> > Using `System Preferences` is easier than changing the settings in plist, and we don't have to do it every time a new version is available. I think you should remove that property.
> 
> Unfortunately, changing just the option in settings doesn't have any outcome.

Have you installed the VoodooI2CHID.kext with the change [here](https://github.com/alexandred/VoodooI2CHID/compare/master...Goshin:ft-prop)?
[Archive.zip](https://github.com/alexandred/VoodooI2C/files/4494194/Archive.zip)

Mateo1234454545:
@Goshin , yes I have. I'm using it now with the option for forcetouch in plist disabled.
And it works licke a charm.
Waiting for you to make a release , for both HS and Catalina.
Thank you for your great work.

alexandred:
@Goshin if you are not able to get your code to grab/set multitouch properties working please see the following: https://github.com/alexandred/VoodooI2C/commit/37febae265c485772e9a480cb0ca086a295a4769#diff-64814a23bc4e5e68123bb895124a1d58R273

At some point I had done the same but it got removed somewhere along the line as it wasn't being used. If i recall correctly, it wasn't enough to set the properties on the HID interface itself, it had to be on the actual IOService object corresponding to the multitouch engine. The code in the above link did that job at least back then.

Mateo1234454545:
> > > `A question, should we keep the configuration property in plist?`
> > > Using `System Preferences` is easier than changing the settings in plist, and we don't have to do it every time a new version is available. I think you should remove that property.
> > 
> > 
> > Unfortunately, changing just the option in settings doesn't have any outcome.
> 
> Have you installed the VoodooI2CHID.kext with the change [here](https://github.com/alexandred/VoodooI2CHID/compare/master...Goshin:ft-prop)?
> [Archive.zip](https://github.com/alexandred/VoodooI2C/files/4494194/Archive.zip)

Correction.
This version works very well even with EnableForceTouchEmulation set to YES.

Goshin:
> @Goshin , yes I have. I'm using it now with the option for forcetouch in plist disabled.
> And it works licke a charm.
> Waiting for you to make a release , for both HS and Catalina.
> Thank you for your great work.

So the configuration property in plist could be a good backup option. Thanks for testing.

@alexandred
Thanks for the info. I have tested changing the type of MultitouchInterface to IOHIDEventService, but the setProperties() didn't get called when properties being changed. Maybe it would also need an HID Device nub or some properties in plist; I don't know, I am new to IOKit.

I don't see the MultitouchPreferences property of VoodooInputSimulatorDevice in IORegExplr; the related properties present in VoodooI2CPrecisionTouchpadHIDEventDriver > HIDEventServiceProperties. but yes we can actively obtain the properties in Multitouch side, without the change in satellites, though we will need a timer to trigger this (maybe 0.5 or 1Hz).

alexandred:
@Goshin The "MultitouchPreferences" property is present on the "AppleMultitouchDevice" entry which is a child of our simulator. This is the reason we have to bubble down the tree to search for it by using the "kIORegistryIterateRecursively" key.

The properties set in sysprefs are precisely the ones in this "MultitouchPreference" dictionary and so for the change to be registered in sysprefs it must be adjusted there (at least, this is what I remember being the case).

Goshin:
Got it, you totally right.

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
```diff
@@ -114,16 +88,57 @@ bool VoodooI2CNativeEngine::start(IOService* provider) {
     setProperty(kIOFBTransformKey, 0ull, 32);
     setProperty("VoodooInputSupported", kOSBooleanTrue);
 
-    isForceClickEnabled = true;
-    isForceTouchEmulationEnabled = true;
-    OSBoolean* enableForceTouchEmulationProperty = OSDynamicCast(OSBoolean, this->getProperty("EnableForceTouchEmulation"));
-    if (enableForceTouchEmulationProperty) {
-        isForceTouchEmulationEnabled = enableForceTouchEmulationProperty->getValue();
+
+    OSDictionary* nameMatch = IOService::serviceMatching("AppleMultitouchDevice");
```
alexandred, line 92: This seems like it could cause undefined behaviour. If more than one multitouch device is connected to the system at the same time (say a touchscreen and a trackpad) then this would only grab the first one to connect. I think it is better to grab the multitouch device that exists as a child of the current instance of the native engine and attach to that.

Mateo1234454545:
@Goshin , I'm trying to build your latest voodooi2c commits but keep getting this error
[Screen Shot 2020-04-18 at 16.18.34.pdf](https://github.com/alexandred/VoodooI2C/files/4496678/Screen.Shot.2020-04-18.at.16.18.34.pdf)

Goshin:
Try `git submodule init` and `git submodule update`.

kprinssu:
@Mateo1234454545 Please read https://voodooi2c.github.io/#Build%20Environment/Build%20Environment.

kprinssu:
@Goshin is this PR stable, as in no more changes?

@lvs1974 can you please test?

I don't have a I2C on hand to test with.

Goshin:
> @Goshin is this PR stable, as in no more changes?

Yes, I believe.

kprinssu:
Great, once someone has confirmed that the changes are working I'll merge it in.

lvs1974:
@kprinssu, @Goshin: I have checked changes in this PR, it works.

- When check-box "Force click and haptic feedback" is unchecked, text selection and dragging works find.
- When this check box is checked, I observe the same old issues.

Thank you for your work!

kprinssu:
Thanks @Goshin and @lvs1974. Merging it in now.

PR #295, opened 2020-04-27
Correct issue with touchscreens requiring an additional finger for mu…

blankmac:
…ltitouch gestures.  This was due to the device presenting a stylus at transducer 0.  A simple check prior to sending the message to VoodooInput eliminates the problem.  The stylus check in VoodooInput should be removed as unnecessary.

### DIFF
--- a/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
+++ b/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
@@ -95,6 +95,8 @@ bool VoodooI2CNativeEngine::start(IOService* provider) {
     
     setProperty(kIOFBTransformKey, 0ull, 32);
     setProperty("VoodooInputSupported", kOSBooleanTrue);
+    
+    stylus_check = 0;
 
     return true;
 }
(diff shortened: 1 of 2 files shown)


blankmac:
Tested and working on my device.

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
```diff
@@ -20,8 +20,16 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
...
+
+    if (!transducer)
+        return MultitouchReturnBreak;
+    
+    if (transducer->type == kDigitiserTransducerStylus)
+        stylus_check = 1;
+    
     for (int i = 0; i < event.contact_count; i++) {
-        VoodooI2CDigitiserTransducer* transducer = (VoodooI2CDigitiserTransducer*) event.transducers->getObject(i);
+        VoodooI2CDigitiserTransducer* transducer = (VoodooI2CDigitiserTransducer*) event.transducers->getObject(i+stylus_check);
```
Goshin, line 32: would be better to reformat the `i+stylus_check` here?

blankmac: It's not clear to me what you are suggesting.  Please elaborate or create a pull request if you feel something is incorrect.

Goshin: I meant change `i+stylus_check` to `i + stylus_check`

blankmac: Oh, I see.  lol.  Sure, if that's the convention that is being used now.  Thanks for pointing it out.

Goshin: It is just my personal suggestion. It seems we don't have a style guideline for this repo.

ben9923: We should embrace the linter we’re using for some of the project code in other parts as a start :)

PR #292, opened 2020-04-22
Override handleIsOpen for MT engine to fix leak

Goshin:
Fix VoodooInput instance leaks when unloading VoodooI2CHID. Refer to https://github.com/alexandred/VoodooI2C/issues/291.

PR #298, opened 2020-04-29
Track interrupt state in I2C controller

Goshin:
Track the interrupt state in the I2C controller to avoid the interrupt being disabled and unregistered twice when unloading.

I notice when kextunloading VoodooI2CHID, the `VoodooI2CControllerDriver::setPowerState` gets called and disables the interrupt. So the interrupt will be disabled and unregistered for the second time in `stop()` when unloading VoodooI2C, which somehow will cause a deadlock in the kernel.

### DIFF
--- a/VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CControllerDriver.cpp
+++ b/VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CControllerDriver.cpp
@@ -357,8 +357,11 @@ void VoodooI2CControllerDriver::readFromBus() {
 }
 
 void VoodooI2CControllerDriver::releaseResources() {
-    nub->disableInterrupt(0);
-    nub->unregisterInterrupt(0);
+    if (is_interrupt_registered) {
+        nub->disableInterrupt(0);
+        nub->unregisterInterrupt(0);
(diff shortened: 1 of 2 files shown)


kprinssu:
This looks good to me as it's mostly refactoring. I am goign to wait on @ben9923 to also take a look.

ben9923:
Looks good :)

tiger511:
Does that mean we can use kextunload and not get a panic?!

kprinssu:
I was investigating that last year. It looks like something was retaining and not properly releasing when a satellite kext was loaded. The panic is probably due to undefined behaviour. 

For now, let's create a Github issue and track it down there instead of commenting on this PR.

Goshin:
>Does that mean we can use kextunload and not get a panic?!

Generally yes, I can now kextunload VoodooI2CHID, VoodooI2C, VoodooGPIO, and VoodooInput, one by one in order.

Some other leaks and crashes have already been fixed in https://github.com/alexandred/VoodooGPIO/pull/4, https://github.com/alexandred/VoodooI2C/pull/292.
Besides, it is still problematic unloading VoodooI2CServices. I have not dug into it yet because currently it is not necessary to reload it when debugging. better to discuss it in a new specific issue.

PR #314, opened 2020-05-14
Add Comet Lake PCI IDs

matthew728960:


kprinssu:
Looks good as it's a simple plist change.

PR #333, opened 2020-06-05
Fixes VoodooInput compatibility 

kprinssu:
Upstream changes in VoodooInput (https://github.com/acidanthera/VoodooInput/commit/835731a0e1547a62b645f4b49b4d3347d3f96dbe) breaks compatibility as the event fields are now changed.

This PR fixes that by assigning an unique finger value to each transducer (done so by transducer id).

### DIFF
--- a/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
+++ b/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
@@ -78,7 +78,11 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
         freeFingerTypes[kMT2FingerTypeUndefined] = false;
     }
     
-    setThumbFingerType(&message);
+    // Fixes 2F taps as MT stack
+    if (event.contact_count > 2) {
+        setThumbFingerType(&message);
+    }
+    
     for(int i = 0; i < event.contact_count; i++) {
         VoodooInputTransducer* inputTransducer = &message.transducers[i];
         
--- a/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
+++ b/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
@@ -25,14 +25,10 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
     if (!transducer)
         return MultitouchReturnBreak;
     
-    if (!event.contact_count) {
-        memset(freeFingerTypes, true, kMT2FingerTypeCount);
-        freeFingerTypes[kMT2FingerTypeUndefined] = false;
-    }
-    
     if (transducer->type == kDigitiserTransducerStylus)
         stylus_check = 1;
     
+    bool allInvalid = true;
     for (int i = 0; i < event.contact_count; i++) {
         VoodooI2CDigitiserTransducer* transducer = (VoodooI2CDigitiserTransducer*) event.transducers->getObject(i+stylus_check);
         VoodooInputTransducer* inputTransducer = &message.transducers[i];
@@ -41,12 +37,13 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
             continue;
         }
         
-        inputTransducer->fingerType = getFingerType(transducer);
+        inputTransducer->fingerType = kMT2FingerTypeUndefined;
         inputTransducer->secondaryId = transducer->secondary_id;
         
         inputTransducer->type = (transducer->type == DigitiserTransducerType::kDigitiserTransducerFinger) ? VoodooInputTransducerType::FINGER : VoodooInputTransducerType::STYLUS;
         
         inputTransducer->isValid = transducer->is_valid;
+        allInvalid = transducer->tip_switch.value() && allInvalid;
         inputTransducer->isTransducerActive = transducer->tip_switch.value();
         inputTransducer->isPhysicalButtonDown = transducer->physical_button.value();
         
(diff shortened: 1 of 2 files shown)


kprinssu:
@ben9923 @Goshin Please take a look.

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
```diff
@@ -70,11 +72,60 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
...
+    
+    // Fixes 2F taps as MT stack
+    if (event.contact_count > 2) {
+        setThumbFingerType(&message);
+    }
+    
+    for(int i = 0; i < event.contact_count; i++) {
+        VoodooInputTransducer* inputTransducer = &message.transducers[i];
+        
+        if (inputTransducer->fingerType == kMT2FingerTypeUndefined) {
```
Goshin, line 89: always true because of the assignment above?
```cpp
inputTransducer->fingerType = kMT2FingerTypeUndefined;
```

It will fail to preserve the previous set finger type.

Goshin:
Unlike VoodooPS2, we should track the finger type of touch in a way based on the `transducer->secondary_id` reported by the device. It is a little bit hard to explain. The finger touches in `event.transducers` are not necessarily in chronological order, so you have to associate the finger type with the corresponding touch based on its identifier id, rather than its ordinal number in the array.

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
kprinssu: @Goshin as I am not familiar with the problem, would you mind taking an attempt at it?

Goshin:
Maybe we don't have to track the finger type info for touche as I mentioned in https://github.com/VoodooI2C/VoodooI2C/issues/332#issuecomment-639889825.

For the change in the message definition of VoodooInput, We can simply port the related part back to VoodooI2C, like https://github.com/Goshin/VoodooI2C/commit/1b910ad2dc2efbd21cd9554a061e8955977ed731, though I think it should be done in MT2Simulator if it is needed by both VoodooPS2 and VoodooI2C.

Btw, I am curious why not add VoodooInput as a git submodule like other dependencies? By doing so we can keep a snapshot of the repo so we won't lose the API compatibility and people won't get a confusing compile error when a header file changes in VoodooInput's master branch.

vit9696:
@Goshin

The change you suggested is rather a hack. I had a brief talk with @usr-sse2, who is a bit occupied at the moment, and the thing is, even with your change VoodooInput will still get finger reenumeration. For example, you release the middle finger and your finger becomes the middle finger. We do not think it is a good and clean idea for compatibility with macOS even if it currently does not fully rely on that. The emulation should really be more or less precise, and with VoodooPS2 we have relatively decent finger enumeration and finger type detection.

As for submodules the issue is updating, as with the submodules you need to make two (or rather three) commits whenever making changes to VoodooInput, which is expensive. We did not really expect that VoodooI2C changes will take long to land, I guess should do that via branches/PRs next time future for better collaboration, sorry.

Goshin:
@vit9696 

> even with your change VoodooInput will still get finger reenumeration. For example, you release the middle finger and your finger becomes the middle finger. We do not think it is a good and clean idea for compatibility with macOS even if it currently does not fully rely on that.

This really happens in real MacBook trackpads according to some collected data. It looks pretty normal to me because the finger type of a touch can change in the middle of a multi-finger gesture when a new touch shows up with higher confidence of that particular finger type. I believe the fingers are not strictly tracked but keep being detected and updated, so it is okay to change the finger type in the midway. Apparently telling apart the rest four fingers doesn't really matter to gesture recognizing at least for now, so I think it is also okay to keep the track of only the thumb.

> The emulation should really be more or less precise, and with VoodooPS2 we have relatively decent finger enumeration and finger type detection.

Cool. Anyway, it is great to have the finger type field in the message. Maybe other clients can perform a more sophisticated detection and inference based on size, shape, pressure, etc.

> As for submodules the issue is updating, as with the submodules you need to make two (or rather three) commits whenever making changes to VoodooInput, which is expensive.

I just roughly look through the discussion in https://github.com/VoodooI2C/VoodooI2C/pull/257. I have no problem with the identity of VoodooInput. It was just a common suggestion to deal with a module with header file dependency.

vit9696:
> I just roughly look through the discussion in #257. I have no problem with the identity of VoodooInput. It was just a common suggestion to deal with a module with header file dependency.

The use of submodules has little to do with identity, it is just that they are very inconvenient for dependency management. The only real issue here is that we broke API compatibility without prior notice.

> This really happens in real MacBook trackpads according to some collected data. It looks pretty normal to me because the finger type of a touch can change in the middle of a multi-finger gesture when a new touch shows up with higher confidence of that particular finger type. I believe the fingers are not strictly tracked but keep being detected and updated, so it is okay to change the finger type in the midway. Apparently telling apart the rest four fingers doesn't really matter to gesture recognizing at least for now, so I think it is also okay to keep the track of only the thumb.

That's very interesting, well, this means that we probably do a little better. But you are right that the clients can be implemented in a more flexible way with the finger type in the message. I believe once this PR lands it will be for the better for all parties involved.

kprinssu:
I will work on it sometime this week. However, I am quite busy until the 15th.

If anybody wants to improve on what I have done so far, please do so. 

@vit9696 it's pretty much just me and @ben9923 maintaining the project. We both have busy lives and can't devote much time VoodooI2C due to other commitments and priorities.

usr-sse2:
> Btw, I am curious why not add VoodooInput as a git submodule like other dependencies? By doing so we can keep a snapshot of the repo so we won't lose the API compatibility and people won't get a confusing compile error when a header file changes in VoodooInput's master branch.

I am both hands for using submodules, and in fact VoodooPS2 was using submodules before @vit9696 rewritten dependency management using custom hacky scripts.

usr-sse2:
> This really happens in real MacBook trackpads according to some collected data. It looks pretty normal to me because the finger type of a touch can change in the middle of a multi-finger gesture when a new touch shows up with higher confidence of that particular finger type. I believe the fingers are not strictly tracked but keep being detected and updated, so it is okay to change the finger type in the midway. Apparently telling apart the rest four fingers doesn't really matter to gesture recognizing at least for now, so I think it is also okay to keep the track of only the thumb.
Yes, but, as you said, finger types should be changed when there is a *higher* confidence, and what we currently have is that when there are index, middle and ring fingers on the touchpad and the middle finger is lifted – we are confident that one of the fingers that is still on the touchpad is the same finger that was previously detected as ring, but it suddenly changes type to middle.

> > As for submodules the issue is updating, as with the submodules you need to make two (or rather three) commits whenever making changes to VoodooInput, which is expensive.
I think that three commits are not as expensive as the issues that happen when VoodooPS2 or VoodooI2C uses an incompatible VoodooInput.

Goshin:
> Yes, but, as you said, finger types should be changed when there is a higher confidence, and what we currently have is that when there are index, middle and ring fingers on the touchpad and the middle finger is lifted – we are confident that one of the fingers that is still on the touchpad is the same finger that was previously detected as ring, but it suddenly changes type to middle.

You are right, but what I meant to say is that finger type changing is allowed in OS gesture handling, and it doesn't care so much about distinguishing the rest four fingers besides thumb. So for the two reasons, I think it is not highly necessary to make efforts to implement the finger tracking just for a rarely used gesture. Tracking the thumb is good enough to solve the original problem well. Please correct me if you find other gestures can also get improved by tracking all fingers.

PR #341, opened 2020-06-15
Update VoodooI2CPCIController for newer Comet Lake PCI IDs

apocolipse:
This hacky patch should probably use something more reliable to determine if its a Comet Lake CPU, the initial check looked for Device ID's starting with `2e` however 2 new device ID's have since been added starting with `6e`.  Updated check here to look for `2e` or `6e`, which got it loading on my 10875H with device ID `6e8`


kprinssu:
Looks good to me, merging it in.

PR #344, opened 2020-07-09
Notify multitouch engine when property changed

Goshin:
- Remove a redundant fixed `IOFBTransform` property in `VoodooI2CNativeEngine`.
- Override `::setProperty()` for `MultitouchInterface` to notify engines on property change.
- Add `onPropertyChange()` method for `MultitouchEngine` to handle property change.

### DIFF
--- a/Multitouch Support/VoodooI2CMultitouchInterface.cpp
+++ b/Multitouch Support/VoodooI2CMultitouchInterface.cpp
@@ -54,6 +54,18 @@ bool VoodooI2CMultitouchInterface::handleIsOpen(const IOService *forClient ) con
     return false;
 }
 
+bool VoodooI2CMultitouchInterface::setProperty(const OSSymbol* key, OSObject* object) {
+    bool result = super::setProperty(key, object);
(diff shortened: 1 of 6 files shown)


Multitouch Support/VoodooI2CMultitouchInterface.cpp
```diff
@@ -54,6 +54,18 @@ bool VoodooI2CMultitouchInterface::handleIsOpen(const IOService *forClient ) con
     return false;
 }
 
+bool VoodooI2CMultitouchInterface::setProperty(const OSSymbol* key, OSObject* object) {
+    bool result = super::setProperty(key, object);
+    if (result) {
```
ben9923, line 59: I usually like to return in case of an error, to avoid excessive indentation, but it's just coding style and we don't have such convention :)

Goshin: Thanks for the advice.

kprinssu:
Thanks @ben9923 and @Goshin.

Issue #342, opened 2020-06-16
Ice Lake Intel I2C


MVDB0110:
Do you have any idea, when Ice Lake will be supported in your kext? I see in the commits that a Ice Lake id is inserted in one of the plists, but there is no support in the kext for the controller.

ben9923:
Which PCI ID does your device have?
If VoodooI2C doesn't have it, can you try manually adding it & installing the kext(s)?

schwxr:
This may not be the case for you, but in my ICL build, VoodooI2C wouldn't load unless I applied the CML fix from #245 to my I2C controller. From there, I was able to proceed with my I2C device as usual.

kprinssu:
@schwxr That fix has been added to upstream since 2.4 (March 2020), see https://github.com/VoodooI2C/VoodooI2C/commit/6d37695082eab4fd9d0203d3a46c7968f9c845d. Is there an underlying issue that has not been fixed?

@MVDB0110 ICL support should've been added with https://github.com/VoodooI2C/VoodooI2C/commit/c3bf644abb510bcc96d553c625ce469e4a2d95aa. As @ben9923 pointed out, do you perhaps have a controller that is not within that list?

KhoraLee:
@kprinssu 6d37695082eab4fd95d0203d3a46c7968f9c845d This only fix CML laptops. ICL(Ice lake) Laptops also need same patch to use I2C Trackpad.

![Screen Shot 2020-08-10 at 6 21 47 PM](https://user-images.githubusercontent.com/18005062/89768275-5c364980-db36-11ea-91d3-73a9e5d17ce5.png)
And pci8086,34e9 is missing.

After fixing two issue 
![Screen Shot 2020-08-10 at 6 09 08 PM](https://user-images.githubusercontent.com/18005062/89768342-73753700-db36-11ea-8782-8d6f1598ae3f.png)
Trackpad works fine

kprinssu:
Thanks @KhoraLee. Merged the changes into master.

KhoraLee:
Now we need to figure out how to check ICL CPUs in order to apply this patch (6d37695082eab4fd95d0203d3a46c7968f9c845d) not only CML but also ICL.

KhoraLee:
@kprinssu Please check this PR. I used same method to verify CPU. 

+It is working for me

kprinssu:
@KhoraLee You want something like this for that if condition:
```
[...]

PR #363, opened 2020-08-10
Update VoodooI2CPCIController for Ice Lake PCI IDs

KhoraLee:
#342

### DIFF
--- a/VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CPCIController.cpp
+++ b/VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CPCIController.cpp
@@ -14,7 +14,7 @@ OSDefineMetaClassAndStructors(VoodooI2CPCIController, VoodooI2CController);
 void VoodooI2CPCIController::configurePCI() {
     char tmp[2];
     const char kCometLakeflag[3] = {'2', '6', 'e'};
-    const char kIceLakeflag[2] = {'3','4'};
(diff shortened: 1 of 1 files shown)


Goshin:
Can it be simplified with `sscanf`?

```c++
int pre, post;
if (sscanf(physical_device.name, "pci8086,%de%d", &pre, &post) == 2) {
    // do the patch here
}
```

KhoraLee:
@Goshin Great Idea!

kprinssu:
Thanks @KhoraLee and @Goshin. Revised code looks much better, except for this one small bug:

`"pci8086,%de%d"` should be `"pci8086,3%de%d".

alexandred:
The PR looks fine to me but is not very modular. Personally I would subclass VoodooI2CPCI for each generation (so VoodooI2CPCICometLake etc) and have the work of checking device ID done by IOKit’s matching process. Each device’s patch would then be isolated to its own PCI subclass. Doing it this way prevents clutter and unnecessary code execution (since these checks are being performed on all systems regardless of their architecture).

kprinssu:
@alexandred Modularising that seems a bit outside the scope of this PR, I suggest that we should address in a subsequent one?

alexandred:
Sure, my only point was that I would favour modularisation instead of merging this PR. The way this is going is that for every future architecture we might need a patch and I’d rather nip this is in the bud now rather than in the future when we have several patches cluttering the code.

That being said I don’t have time to do this myself so I am happy to merge this PR to provide the desired functionality asap.

PR #365, opened 2020-08-11
Attempt to get GPIO Interrupt from _DSM

zhen-zen:
For ACPI code of some I2C controller, GPIO interrupt are also available from `_DSM` with an undocumented GUID `ef87eb82-f951-46da-84ec-14871ac6f84b`. When both APIC and GPIO are unavailable, check `_DSM` for GPIO.

Sample ASL code:
```
Method (TP7D, 6, Serialized)
{
    If ((Arg0 == TP7G))
    {
        If ((Arg2 == Zero))
        {
            If ((Arg1 == One))
            {
                Return (Buffer (One)
                {
                     0x03                                           
                })
            }
        }

        If ((Arg2 == One))
        {
            Return (ConcatenateResTemplate (Arg4, Arg5))
        }
    }

    Return (Buffer (One)
    {
         0x00                                           
    })
}

Method (_DSM, 4, Serialized)  // _DSM: Device-Specific Method
{
    If ((Arg0 == HIDG))
    {
        Return (HIDD (Arg0, Arg1, Arg2, Arg3, HID2))
    }

    If ((Arg0 == TP7G))
    {
        Return (TP7D (Arg0, Arg1, Arg2, Arg3, SBFB, SBFG))
    }

    Return (Buffer (One)
    {
         0x00                                           
    })
}

```

### DIFF
--- a/VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
+++ b/VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
@@ -116,7 +116,7 @@ IOReturn VoodooI2CDeviceNub::getDeviceResourcesDSM(UInt32 index, OSObject **resu
     data->release();
 
     if (!(availableIndex & (index << 1))) {
-        IOLog("%s::%s TP7D index 0x%x doesn't support GPIO report\n", controller_name, getName(), availableIndex);
+        IOLog("%s::%s TP7D doesn't support index 0x%x\n", controller_name, getName(), availableIndex);
         return kIOReturnUnsupportedMode;
     }
 
@@ -126,7 +126,7 @@ IOReturn VoodooI2CDeviceNub::getDeviceResourcesDSM(UInt32 index, OSObject **resu
 bool VoodooI2CDeviceNub::getAlternativeGPIOInterrupt(VoodooI2CACPICRSParser* crs_parser) {
     OSObject *result;
     if (getDeviceResourcesDSM(TP7G_GPIO_INDEX, &result) != kIOReturnSuccess) {
-        IOLog("%s::%s Warning: If your chosen satellite implements polling then %s will run in polling mode.\n", controller_name, getName(), getName());
+        IOLog("%s::%s failed to retrieve GPIO interrupt from _DSM or XDSM\n", controller_name, getName());
         result->release();
         return false;
     }
@@ -147,7 +147,6 @@ bool VoodooI2CDeviceNub::getAlternativeGPIOInterrupt(VoodooI2CACPICRSParser* crs
         return false;
     }
 
-    has_gpio_interrupts = true;
     IOLog("%s::%s GPIO interrupts retrieved through _DSM method", controller_name, getName());
     return true;
 }
--- a/VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
+++ b/VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
@@ -38,10 +38,6 @@ bool VoodooI2CDeviceNub::attach(IOService* provider, IOService* child) {
         return false;
     }
 
-    if (getHIDDescriptorAddress() != kIOReturnSuccess) {
-        IOLog("%s::%s Could not get device HID descriptor\n", controller_name, child->getName());
-    }
-
     if (has_gpio_interrupts) {
         gpio_controller = getGPIOController();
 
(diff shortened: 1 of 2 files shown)


Goshin:
You can extract the common part of invoking `_DSM` to a function like 
`IOReturn evaluateDSM(uuid_t guid, OSObject *result)`. It will be also available for the HID satellite.

And I would merge the part of getting the GPIO pin from DSM into here:
https://github.com/VoodooI2C/VoodooI2C/blob/3527ec36d2f5860253544f39bec6f0998a7044e2/VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp#L102-L111

Goshin:
A simpler version:

```diff
@@ -99,7 +99,7 @@ IOReturn VoodooI2CDeviceNub::getDeviceResources() {
         return kIOReturnNotFound;
     }

-    if (crs_parser.found_gpio_interrupts) {
+    if (crs_parser.found_gpio_interrupts || (getDeviceResourcesDSM() == kIOReturnSuccess && getAlternativeGPIOInterrupt(&crs_parser) == kIOReturnSuccess)) {
         setProperty("gpioPin", crs_parser.gpio_interrupts.pin_number, 16);
         setProperty("gpioIRQ", crs_parser.gpio_interrupts.irq_type, 16);

         has_gpio_interrupts = true;
         gpio_pin = crs_parser.gpio_interrupts.pin_number;
         gpio_irq = crs_parser.gpio_interrupts.irq_type;
     } else {
         has_gpio_interrupts = false;
     }
```

zhen-zen:
It seems that native APIC will never be used if alternative GPIO is retrieved early?

Goshin:
Yes, I don't think it has to be the last process. It and evaluating `_CRS` are both for getting GPIO config, so we can put them together.

Btw `getDeviceResourcesDSM()` can be placed inside `getAlternativeGPIOInterrupt()`.

zhen-zen:
Oh I got confused with APIC interrupts. So it's only retrieved from `gIOInterruptSpecifiersKey` but not parsed in `_CRS`. Then GPIO config will be evaluated regardless of whether APIC interrupt is available?

Goshin:
APIC interrupts are handled by IOKit. Those checks are to ensure the APIC pin is compatible.

zhen-zen:
> A simpler version:
> 
> ```diff
> @@ -99,7 +99,7 @@ IOReturn VoodooI2CDeviceNub::getDeviceResources() {
>          return kIOReturnNotFound;
>      }
> 
> -    if (crs_parser.found_gpio_interrupts) {
> +    if (crs_parser.found_gpio_interrupts || (getDeviceResourcesDSM() == kIOReturnSuccess && getAlternativeGPIOInterrupt(&crs_parser) == kIOReturnSuccess)) {
>          setProperty("gpioPin", crs_parser.gpio_interrupts.pin_number, 16);
>          setProperty("gpioIRQ", crs_parser.gpio_interrupts.irq_type, 16);
> 
>          has_gpio_interrupts = true;
>          gpio_pin = crs_parser.gpio_interrupts.pin_number;
>          gpio_irq = crs_parser.gpio_interrupts.irq_type;
>      } else {
>          has_gpio_interrupts = false;
>      }
> 
> 
>      if (!has_gpio_interrupts) {
>          OSArray* interrupt_array = OSDynamicCast(OSArray, acpi_device->getProperty(gIOInterruptSpecifiersKey));
>          if (!interrupt_array) {
>              IOLog("%s::%s Warning: Could not find any APIC nor GPIO interrupts; if your chosen satellite implements polling then %s will run in polling mode.\n", controller_name, getName(), getName());
>              goto exit;
>          }
> 
>      ...
> ```

By adding getAlternativeGPIOInterrupt() at this stage, it will override possible APIC interrupt if available.

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
```diff
@@ -72,20 +71,120 @@ IOReturn VoodooI2CDeviceNub::enableInterrupt(int source) {
...
+        OSNumber::withNumber(I2C_DSM_REVISION, 32),
+        OSNumber::withNumber(index, 32),
+        OSArray::withCapacity(1),
+    };
+
+    ret = acpi_device->evaluateObject("_DSM", result, params, 4);
+    if (ret != kIOReturnSuccess)
+        ret = acpi_device->evaluateObject("XDSM", result, params, 4);
+
+    params[0]->release();
```
ben9923, line 95: They can be freed in a loop instead. Doesn't matter much anyway...

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
```diff
@@ -72,20 +71,120 @@ IOReturn VoodooI2CDeviceNub::enableInterrupt(int source) {
...
+    // convert to mixed-endian
+    *(reinterpret_cast<uint32_t *>(guid)) = OSSwapInt32(*(reinterpret_cast<uint32_t *>(guid)));
+    *(reinterpret_cast<uint16_t *>(guid) + 2) = OSSwapInt16(*(reinterpret_cast<uint16_t *>(guid) + 2));
+    *(reinterpret_cast<uint16_t *>(guid) + 3) = OSSwapInt16(*(reinterpret_cast<uint16_t *>(guid) + 3));
+
+    OSObject *params[] = {
+        OSData::withBytes(guid, 16),
+        OSNumber::withNumber(I2C_DSM_REVISION, 32),
+        OSNumber::withNumber(index, 32),
+        OSArray::withCapacity(1),
```
ben9923, line 88: Finally getting rid of this warning!
Wanted to do it for a long time, but don't have an HID device :P

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
```diff
@@ -72,20 +71,120 @@ IOReturn VoodooI2CDeviceNub::enableInterrupt(int source) {
...
+    }
+
+    return evaluateDSM(I2C_DSM_TP7G, index, result);
+}
+
+IOReturn VoodooI2CDeviceNub::getAlternativeInterrupt(VoodooI2CACPICRSParser* crs_parser) {
+    OSObject *result = nullptr;
+    if (getDeviceResourcesDSM(TP7G_GPIO_INDEX, &result) != kIOReturnSuccess) {
+        IOLog("%s::%s failed to retrieve interrupts from _DSM or XDSM method\n", controller_name, getName());
+        result->release();
```
ben9923, line 130: You probably want to use `OSSafeReleaseNULL` here.
As it's in an error-handling context, there's a high possibility of `result` being `nullptr`.

You might want to enforce the usage of this macro in other places too. It also saves a manual check for null, and prevents accidental undefined behaviors.

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
```diff
@@ -72,20 +71,120 @@ IOReturn VoodooI2CDeviceNub::enableInterrupt(int source) {
...
+        return kIOReturnNotFound;
+    }
+
+    OSData *data = OSDynamicCast(OSData, result);
+    if (!data) {
+        IOLog("%s::%s Could not get valid return for interrupts from _DSM or XDSM method\n", controller_name, getName());
+        result->release();
```
ben9923, line 137: Same note about safe release. I'd use the macro instead of all of the calls to `release()`.

ben9923 (changes requested):
Thanks for the PR!
Wish we could put it in VoodooI2CHID, as it's HID-specific code, but the current driver design won't allow it.

Note for the future - We should probably prefer APIC interrupts if possible. We currently prefer GPIO... (But let's keep it out of this PR's scope).

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
zhen-zen: Actually I wonder if if `_DSM` to `XDSM` patch is still being used today. IIRC addProperty approach is already being adopted at early stage of VoodooI2C development.

PR #385, opened 2020-10-03
Follow up of requested changes in #365; potential optimization of APIC / GPIO interrupt validation

zhen-zen:
https://github.com/VoodooI2C/VoodooI2C/pull/365#pullrequestreview-501553472

@Goshin @ben9923 @kprinssu 

Thank you for your review on the previous PR.

Current checklist for device resource is:
1. evaluate `_CRS`
This method must exist and allow custom modification (`SBFI` -> `SBFG` etc)
I2C declaration is obtained here
2. check for GPIO interrupt
It's either a native one or modified by user (who take responsibility for valid pin config)
No additional check here, skip next step
3. If APIC interrupt pin is not valid, then evaluate `_DSM` for alternative GPIO
Corresponding `_DSM` GUID may not exist in some early models.
4. If GPIO interrupt is selected, get pin and irq. Otherwise, APIC interrupt will just work or try polling.

If APIC is to be prioritized, APIC will be verified first and other information in `_CRS` will be ignored. I suppose user won't have choice to override that?


kprinssu:
This looks good to me.

kprinssu (approved):

Issue #38, opened 2017-05-24
Gitter Broke


LiteGaming:
I can't get into the chat room for some reason?

alexandred:
You were temporarily banned for off-topic disussion. I have lifted the ban. Please stick to discussion about VoodooI2C in the future.

PR #390, opened 2020-10-06
Catch another garbage pointer

zhen-zen:


hieplpvip:
I think `getDeviceResourcesDSM` should have its own `result`. The kernel panic has been caused because of forgetting to clear the pointer, which shouldn't happen at all if `getDeviceResourcesDSM` uses its own `result`.

A little off-topic. @kprinssu Could you please be more careful when merging pull requests? @zhen-zen said my pull request (https://github.com/VoodooI2C/VoodooI2C/pull/386) was a duplicate of https://github.com/VoodooI2C/VoodooI2C/pull/385, so I close my pull request. It's my bad that I didn't take a close look at that pull request to discover that it doesn't fix the kernel panic. The point is, you merged it and released a new version (v2.5.1) that actually doesn't fix anything.

It took me a while to figure out what caused the kernel panic, only to create a pull request then close it because I thought it had been fixed in another one, which isn't true. I'm somewhat disappointed now.

I don't mean any disrespect. You guys have done an excellent job for the Hackintosh community. Just please be more careful when merging pull requests.

zhen-zen:
We were both too hurry to get it fixed.

I agree with a separate OSObject result, which is more fault tolerant since the release will only be executed on one `executeDSM`.

And for your PR, if the return result is not `kIOReturnSuccess`, nothing will be release. Under some circumstance, it may leave unmanaged OSObject. And mine approach is also vulnerable to the dirty pointer.

zhen-zen:
And I didn't mean to offend, can I ask you to also check open PR after find some issues? Which may save a lot of time for debugging. From my side, it's somewhat pushed tightly as well. Such as the string comparison in #37 #38 . Also I was considering to revert previous OC doc change and hold the release for a potential flaw in `BrightnessKey`. And you just edited the doc...

zhen-zen:
Got some more samples, currently testing different `_DSM` layouts.

zhen-zen:
I just removed the `TP7G` functionality in original `_DSM` and forced a device into polling mode. Maybe I will add it to the document.

```

    Scope (_SB.PCI0.I2C3.SHUB)
    {
        Name (TP7G, ToUUID ("ef87eb82-0000-0000-0000-14871ac6f84b"))
    }

```

Set `*result` to `nullptr` is the last piece of work. My initial thought for the util methods was to have all `OSObject` released in where they are created, thus none of them will be left behind. However, most users I reached for testing (about a dozen?) are are new ICL/CML owners. They haven't developed their I2C patches, so they has the motivation, as well as the `TP7G` support in `_DSM`. So the problem is unlikely to be caught until being merged, even if the double release one is caught before. Unfortunately, that caused the issue we faced now. Only a new release could get broader response as well as the issue underneath.

kprinssu:
@hieplpvip That is my bad and I will own up to it.  PR#385 looked to be good on my end and I did not test as I do not have a device on hand to test with.

For now, I have pulled 2.5.1 and will be waiting on @ben9923 to further review PRs.

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
```diff
@@ -114,6 +114,7 @@ IOReturn VoodooI2CDeviceNub::getDeviceResourcesDSM(UInt32 index, OSObject **resu
 
     UInt8 availableIndex = *(reinterpret_cast<UInt8 const*>(data->getBytesNoCopy()));
     data->release();
+    *result = nullptr;
 
     if (!(availableIndex & (index << 1))) {
```
ben9923, line 119: While somewhat unrelated to this PR, shouldn't this be `1 << index`?

ben9923 (commented):
Looks good to me, as long as it's properly tested...
Check out the `index`-related comment though.

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
zhen-zen: Many thanks, almost overlooked that!

hieplpvip:
Can confirm the kernel panic is fixed

kprinssu:
Thanks @hieplpvip.

LeeBinder:
> Can confirm the kernel panic is fixed

I confirm that, too, also on a Asus Laptop with ELAN TP. Good job!

mishurov:
> Can confirm the kernel panic is fixed

I confirm that too. Asus UX310UQK, ELAN1200. Thank you!

PR #391, opened 2020-10-09
Evaluate XDSM first for easy customization and document updates

zhen-zen:
When drafting prior PR, I have considered if `XDSM` support could be dropped since the approach is already deprecated. However, I found it could be difficult to edit `_DSM` if one want to force polling mode instead of GPIO when updating the docs. I also wonder if GPIO pin would need to be updated under rare circumstances. Then insert `XDSM` could be an easy implementation for customization.

### DIFF
--- a/Documentation/SSDT-I2CD.dsl
+++ b/Documentation/SSDT-I2CD.dsl
@@ -1,7 +1,7 @@
 /*
  * Table for debuging common I2C constants.
  *
- * Please note you might need a remove _INI hacks temporary to reveal
+ * Please note you might need to remove _INI hacks temporary for
  * original value.
  */
 DefinitionBlock ("", "SSDT", 2, "hack", "I2CD", 0x00000000)
@@ -16,6 +16,8 @@ DefinitionBlock ("", "SSDT", 2, "hack", "I2CD", 0x00000000)
     External (SMD2, FieldUnitObj)
     External (SMD3, FieldUnitObj)
     // I2C Interrupt mode
+    External (GPDI, FieldUnitObj)
+    External (GPLI, FieldUnitObj)
     External (SDM0, FieldUnitObj)
     External (SDM1, FieldUnitObj)
     // I2C Device type
@@ -26,14 +28,14 @@ DefinitionBlock ("", "SSDT", 2, "hack", "I2CD", 0x00000000)
     {
         Device (I2CD)
         {
-            Name (_HID, EisaId ("PNP0C02"))  // _HID: Hardware ID
+            Name (_HID, EisaId ("PNP0C02"))
             Name (_STA, 0x0F)  // _STA: Status
 
             If (CondRefOf (\GPEN))
             {
                 Device (GPEN)
                 {
-                    Method (_ADR, 0, Serialized)  // _ADR: Address
+                    Method (_ADR, 0, Serialized)
                     {
                         Return (\GPEN)
                     }
@@ -44,7 +46,7 @@ DefinitionBlock ("", "SSDT", 2, "hack", "I2CD", 0x00000000)
             {
                 Device (SBRG)
                 {
-                    Method (_ADR, 0, Serialized)  // _ADR: Address
+                    Method (_ADR, 0, Serialized)
                     {
                         Return (\SBRG)
                     }
@@ -55,7 +57,7 @@ DefinitionBlock ("", "SSDT", 2, "hack", "I2CD", 0x00000000)
             {
                 Device (SMD0)
                 {
-                    Method (_ADR, 0, Serialized)  // _ADR: Address
+                    Method (_ADR, 0, Serialized)
                     {
                         Return (\SMD0)
                     }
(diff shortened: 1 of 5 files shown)


ffosilva:
I totally agree with this pull request. After 2.5.x upgrade, the automatic detection of GPIO based on _DSM broke my ACPI patching strategy that I used to make my trackpad working on both Linux & macOS booting from OpenCore. If I do the GPIO pinning on _DSM, the trackpad stops working on Linux.

zhen-zen:
> I totally agree with this pull request. After 2.5.x upgrade, the automatic detection of GPIO based on _DSM broke my ACPI patching strategy that I used to make my trackpad working on both Linux & macOS booting from OpenCore. If I do the GPIO pinning on _DSM, the trackpad stops working on Linux.

That's weird since the only evaluation to `_DSM` in linux is for the HID descriptor at `https://github.com/torvalds/linux/blob/master/drivers/hid/i2c-hid/i2c-hid-core.c`. And I'm not quite sure what did you patch. Typically nothing's required for Linux as well as VoodooI2C 2.5.2.

As a reference, you may try following patch.

```

    Scope (_SB.PCI0.I2C3.SHUB)
    {
        Method (TP7G, 0, Serialized)
        {
            If (_OSI("Darwin")) {
                return (ToUUID ("ef87eb82-0000-0000-0000-14871ac6f84b")) // Invalid UUID
            } Else {
                return (ToUUID ("ef87eb82-f951-46da-84ec-14871ac6f84b")) // Original UUID
            }
        }
    }

```

ben9923:
I believe we shouldn't support `XDSM` at all.
I guess it was right to do when a lot of users renamed their `_DSM` methods - Same approach (Using `XDSM` as a fallback, not the other way) was made in RehabMan's kexts back then.

Now when it's not commonly used, and as `XDSM` is just a custom method, I don't think we should rely on it, but rather only rely on methods defined by the ACPI spec & exist natively.

I think the first step would be adding a deprecation message for a while, before removing support for it altogether.

Regarding the ease of patching `_DSM` - It's pretty much a Find/Replace of the `_DSM` name with a couple of extra leading/trailing bytes to only match the wanted method (Easily found using `iasl -l`).
I don't believe a case of a user changing the device's GPIO pin is likely, and would probably discourage any case of using polling mode - cases of non-functional GPIO are much less likely than before (Invalid GPIO configurations, as we've seen in the past days, shouldn't be something we need to support).

If we _are_ going to deprecate `XDSM`, I guess it wouldn't hurt much to put it before `_DSM` evaluation, just as a temporary mean until user's configuration is aligned with the spec...

@ffosilva I, too, couldn't really understand why it broke your configuration. I do suspect it's been broken before the update and just showed itself after it, like some other users.

CC @kprinssu

kprinssu (commented):
This is more stylistic, but can you please wrap the if/else in curly braces (`{`, `}`).

kprinssu:
I also agree with @ben9923, we should not be relying on XDSM as it's a hack/crutch (it's not part of the ACPI specification) and should be deprecated.

ffosilva:
Hi @ben9923. I got your point and totally agree. I reviewed my method and now I don't rely on XDSM patch - now I got my Dell ALPS touchpad working on 2.5.2 by replacing Windows XP -> Darwin for _OSI and changing OSYS value to the same of Windows 10; also made some workaround on _CSR via SSDT instead of changing it directly on DSDT.

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
```diff
@@ -88,9 +88,12 @@ IOReturn VoodooI2CDeviceNub::evaluateDSM(const char *uuid, UInt32 index, OSObjec
         OSArray::withCapacity(1),
     };
 
-    ret = acpi_device->evaluateObject("_DSM", result, params, 4);
-    if (ret != kIOReturnSuccess)
-        ret = acpi_device->evaluateObject("XDSM", result, params, 4);
+    ret = acpi_device->evaluateObject("XDSM", result, params, 4);
+    if (ret != kIOReturnSuccess) {
+        ret = acpi_device->evaluateObject("_DSM", result, params, 4);
```
kprinssu, line 93: You should add a check for `ret` again after Line 93:
```
ret = acpi_device->evaluateObject("_DSM", result, params, 4);
if (ret != kIOReturnSuccess) {
....
}
```

kprinssu (changes requested):
Minor change as the ACPI call may fail.

Otherwise, looks good!

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
zhen-zen: `evaluateDSM` is drafted as a wrapper for `evaluateObject`. The `IOReturn` result will be handled by methods calling it, which is `getDeviceResourcesDSM` in this case. So it's not necessary to verify it here.

kprinssu (approved):

Documentation/SSDT-I2CD.dsl
```diff
@@ -0,0 +1,168 @@
+/*
```
ben9923, line 1: I love it!

ben9923 (approved):
Looks good!

ben9923:
@zhen-zen I recall you had an up-to-date version of the SSDT with some more variables in it (i.e. `TPDM`). Can you please open another PR with those included?

Thanks!

PR #398, opened 2020-11-05
Option to opt-out from _DSM/XDSM interrupts

black-dragon74:
First off, kudos to @zhen-zen for #365 (interesting find). 

However this feature breaks polling mode on select machines wherein the OEM GPIO implementation is buggy, like on most of the Asus laptops.

This PR adds a boot-arg (`-vi2c-no-alt-interrupts`) to opt out from the alternative interrupts.

Regards

### DIFF
--- a/VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
+++ b/VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
@@ -32,14 +32,6 @@ bool VoodooI2CDeviceNub::attach(IOService* provider, IOService* child) {
         return false;
     }
 
-    pci_device = controller->nub->controller->physical_device.pci_device;
-
-    if (checkKernelArg("-vi2c-force-polling")) {
-        force_polling = true;
-    } else if (pci_device) {
-        force_polling = pci_device->getProperty("force-polling") != nullptr;
-    }
-
     if (getDeviceResources() != kIOReturnSuccess) {
         IOLog("%s::%s Could not get device resources\n", controller_name, child->getName());
         return false;
@@ -201,6 +193,17 @@ IOReturn VoodooI2CDeviceNub::getDeviceResources() {
         return kIOReturnNotFound;
     }
 
+    IOPCIDevice *pci_device { nullptr };
+    bool force_polling { false };
+
+    pci_device = controller->nub->controller->physical_device.pci_device;
+
+    if (checkKernelArg("-vi2c-force-polling")) {
+        force_polling = true;
+    } else if (pci_device) {
+        force_polling = pci_device->getProperty("force-polling") != nullptr;
+    }
+
     if (!force_polling) {
         if (crs_parser.found_gpio_interrupts ||
         (!validateInterrupt() && getAlternativeInterrupt(&crs_parser) == kIOReturnSuccess)) {
@@ -212,11 +215,11 @@ IOReturn VoodooI2CDeviceNub::getDeviceResources() {
         gpio_irq = crs_parser.gpio_interrupts.irq_type;
         }
     } else {
-        IOLog("%s::%s Forced polling mode, skipping APIC/GPIO interrupts", controller_name, getName());
+        IOLog("%s::%s Forced polling mode, skipping APIC/GPIO interrupts\n", controller_name, getName());
     }
 
     if (!has_apic_interrupts && !has_gpio_interrupts && !force_polling)
-        IOLog("%s::%s Warning: Could not find any APIC nor GPIO interrupts. Your chosen satellite will run in polling mode if implemented.", controller_name, getName());
+        IOLog("%s::%s Warning: Could not find any APIC nor GPIO interrupts. Your chosen satellite will run in polling mode if implemented.\n", controller_name, getName());
 
     return kIOReturnSuccess;
 }
--- a/VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
+++ b/VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
@@ -196,7 +196,7 @@ IOReturn VoodooI2CDeviceNub::getDeviceResources() {
     uint32_t checkForAltInterrupts;
     if (crs_parser.found_gpio_interrupts ||
         (!validateInterrupt() && getAlternativeInterrupt(&crs_parser) == kIOReturnSuccess &&
-         !PE_parse_boot_argn("-vi2c-no-alt-interrupts", &checkForAltInterrupts, sizeof(checkForAltInterrupts)))) {
+         !PE_parse_boot_argn("-vi2c-force-polling", &checkForAltInterrupts, sizeof(checkForAltInterrupts)))) {
         setProperty("gpioPin", crs_parser.gpio_interrupts.pin_number, 16);
         setProperty("gpioIRQ", crs_parser.gpio_interrupts.irq_type, 16);
 
--- a/VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
+++ b/VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
@@ -31,6 +31,22 @@ bool VoodooI2CDeviceNub::attach(IOService* provider, IOService* child) {
         IOLog("%s::%s Could not get controller\n", controller_name, child->getName());
         return false;
     }
+    
+    auto matchingDict = IOService::serviceMatching("VoodooI2CPCIController");
+    auto service = IOService::waitForMatchingService(matchingDict, 100000000);
+    matchingDict->release();
+    
+    if (service != nullptr) {
+        if (service->getParentEntry(gIOServicePlane) != nullptr) {
+            pci_controller_entry = service->getParentEntry(gIOServicePlane);
+        } else {
+            IOLog("%s::%s Could not get parent PCI controller\n", controller_name, child->getName());
+            return false;
+        }
+    } else {
+        IOLog("%s::%s Could not get VoodooI2CPCIController instance\n", controller_name, child->getName());
+        return false;
+    }
 
     if (getDeviceResources() != kIOReturnSuccess) {
         IOLog("%s::%s Could not get device resources\n", controller_name, child->getName());
@@ -192,11 +208,18 @@ IOReturn VoodooI2CDeviceNub::getDeviceResources() {
         IOLog("%s::%s Could not find an I2C Serial Bus declaration\n", controller_name, getName());
         return kIOReturnNotFound;
     }
-
-    uint32_t checkForAltInterrupts;
+    
+    // Check if user wants to disable alt-interrupts
+    uint32_t forcePolling;
+    if (checkKernelArg("-vi2c-force-polling")) {
+        use_alt_interrupts = false;
+    } else if (readProperty(pci_controller_entry, "force-polling", forcePolling)) {
+        use_alt_interrupts = forcePolling != 1;
+    }
+    
     if (crs_parser.found_gpio_interrupts ||
         (!validateInterrupt() && getAlternativeInterrupt(&crs_parser) == kIOReturnSuccess &&
-         !PE_parse_boot_argn("-vi2c-force-polling", &checkForAltInterrupts, sizeof(checkForAltInterrupts)))) {
+         use_alt_interrupts)) {
         setProperty("gpioPin", crs_parser.gpio_interrupts.pin_number, 16);
         setProperty("gpioIRQ", crs_parser.gpio_interrupts.irq_type, 16);
 
--- a/VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
+++ b/VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
@@ -193,8 +193,10 @@ IOReturn VoodooI2CDeviceNub::getDeviceResources() {
         return kIOReturnNotFound;
     }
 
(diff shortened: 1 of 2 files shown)


zhen-zen:
Hmmm, I'm editing `Polling Mode.md` and will mention this there, including a SSDT sample to break the alternative mechanism. So after determined whether GPIO interrupt mode is usable, it's not necessary to keep the `boot-arg` there.

black-dragon74:
Well I know of the changing `TP7G` UUID but that wouldn't be viable as we should totally avoid messing with ACPI where possible. ACPI approach would require renaming of original `TP7G` and then injecting a new one via `SSDT-I2C`. Too much work. Anyways that's my take on it.

Regards

zhen-zen:
There's no need to rename the `TP7G` under upper scope, just insert an invalid one at `TPDX`/`TPLX` is enough.

black-dragon74:
AFAIK, To stop `_DSM` interrupts, you would have to return an invalid `UUID` from `TP7G` name or method. And we can't have same declaration twice in ACPI otherwise we will get `AE_ALREADY_EXISTS`. Hence to define `TP7G` again as an ACPI `Method` or `Name` object, we will be required to rename the original declaration to something else.

Sorry if I didn't understand what you meant. Could you show by an example?

Regards

black-dragon74:
> Edit: Maybe rename it something like force polling?

Yeah. Sure. I am not good at naming things LOL.

zhen-zen:
Yes, you can check https://github.com/acidanthera/OpenCorePkg/pull/92#issuecomment-673310377.
And there are some people who want a clean `boot-arg`. For example, people tend to replace `alcid=X` with `layout-id` in properties. And a simple name declaration is enough, which can be combined into XOSI ones.

black-dragon74:
> Yes, you can check [acidanthera/OpenCorePkg#92 (comment)](https://github.com/acidanthera/OpenCorePkg/pull/92#issuecomment-673310377).

Got to learn something new. Thanks.

> And there are some people who want a clean `boot-arg`. For example, people tend to replace `alcid=X` with `layout-id` in properties. And a simple name declaration is enough, which can be combined into XOSI ones.

Agreed.

zhen-zen:
When reading `Polling Mode.md`, I was wondering whether to disable this mechanism by default. I admit it was aggressive to enable this feature by default. At that time, I'm not aware of such buggy GPIO pin issues. Users also live with polling mode which is not that bad.

Pros:
Polling mode should always work. (?)
No extra efforts for polling mode users with buggy GPIO interrupts.
Cons:
Users need additional setup to take advantage of this mechanism. Since customized ACPI name is not favorable, maybe something like adding property via bootloader?
```
<key>PciRoot(0x0)/Pci(0x1f,0x3)</key>
<dict>
        <key>force-polling</key>
        <data>AQAAAA==</data>
        <key>gpio-alt</key>
        <data>AQAAAA==</data>
</dict>
```

CC @ben9923 @kprinssu @Goshin what do you think?

black-dragon74:
You know what, actually I thought of same initially but then I saw your PR discussion and figured the team might have thought over this. We should definitely have an option to force polling mode and then an optional config to enable alt-interrupts.

And yeah, device properties is the best way to go. Although, we should also have an equivalent boot-arg for pre install scenarios.

P.S: I actually am running this variant (force polling) on my machine. So if need be, I can push those commits to this branch.

Regards

black-dragon74:
So it just occurred to me, we can't really force polling mode out of the box as polling mode also requires some minor `DSDT` edits. Most of systems that I have seen return `SBFG` from the `_CRS` which would cause polling mode to fail even when the GPIO controller is not available (`_STA` not patched for macOS).

So the best case scenario would be, leaving the current implementation as is with an option to opt out from alternative interrupts using boot-arg/device-props on buggy systems with boot-arg having precedence over device-properties (for macOS installer/recovery)

I personally would not want users to go through the ACPI `TP7G` override as majority of them struggle at ACPI editing but having this as an option would not require any changes in the `VoodooI2C` source code so we can mention about this in the `PollingMode.md`

What are your views on it?

Regards

black-dragon74:
Okay so I did a bit if research and found out that in most of the cases editing `_CRS` is required as there is often code that we don't require when running macOS. Also, I have noticed that most systems return `SBFG` by default, which again thwarts out attempt of enabling polling mode out of the box and if we write the implementation to ignore parsing of device resources when user wants to force polling mode, it still won't work as we want `_CRS` to return `SBFB` for polling mode.

So, if we will be editing ACPI anyways, why not leave the current implementation as is with an option to opt out from alternative interrupts on systems with buggy GPIO implementation. This can be achieved with boot-arg/device-props with boot-arg having the precedence (for macOS installer/recovery).

We can additionally mention the the ACPI way by using `TP7G` override in `PollingMode.md` for a cleaner approach.

Regards

Mateo1234454545:
> You know what, actually I thought of same initially but then I saw your PR discussion and figured the team might have thought over this. We should definitely have an option to force polling mode and then an optional config to enable alt-interrupts.
> 
> And yeah, device properties is the best way to go. Although, we should also have an equivalent boot-arg for pre install scenarios.
> 
> P.S: I actually am running this variant (force polling) on my machine. So if need be, I can push those commits to this branch.
> 
> Regards

Could you send your variant to try it on my laptop?
Thanks in advance.

kprinssu:
@Mateo1234454545 Please only reply if you have something meaningful to contribute to the discussion.

zhen-zen:
Is it necessary to add additional device match? And the driver could even match to other I2C device.
The PCI device should be reachable all the way up like `controller->nub->controller->physical_device` VoodooI2CDeviceNub/VoodooI2CControllerDriver/VoodooI2CControllerNub/VoodooI2CController/VoodooI2CControllerPhysicalDevice

zhen-zen:
Also something to add:
PCI device may be not available on older systems with `VoodooI2CACPIController` instead of `VoodooI2CPCIController`. So it could be like
```
// check boot-arg first
    IOService *pci = controller->nub->controller->physical_device.pci_device;
    if (pci) {
        if (pci->getProperty("force-polling")) {
            force_polling = true;
        }
        if (pci->getProperty("gpio-alt"))
            IOLog("%s::%s gpio-alt!\n", controller_name, getName());
    }
```
And we don't need to know the exact value of that property. It's easier to just remove the entry or rename it to `#force-polling` instead of changing the base64 value to 0.

black-dragon74:
> Is it necessary to add additional device match? 

I quite honestly missed the alternate approach. Thanks. Saves from an additional delay.

> So it sounds like force polling actually does the same thing as no gpio alt mode

Yes it did so as we had not decided which path we wanted to take. We were still discussing it so I skipped that implementation. However, I am pushing the changes now. 

According to current implementation, `VoodooI2CDeviceNub::attach` checks for the boot-arg `-vi2c-force-polling`. If it is not found, it checks for  a `force-polling` property on `I2C` device (regardless of its value). If any of the above two conditions are true, all the GPIO/APIC interrupts are ignored and driver is forced into polling mode.

Also, nothing related to `no-gpio-alt` is implemented as people would only want to use that in cases where the OEM GPIO implementation is buggy. Better use `force-polling` in such cases? Saves the day, right? If not, I can add it too. Just let me know what you think of it.

Regards

zhen-zen:
Cool, I think it's enough here. With `force-polling`, it could either be APIC interrupt mode or just the pulling one. I assume there's should be nothing wrong with APIC one. Maybe only developers of satellite kexts will need to override APIC one for testing polling mode and it's easy to comment out `interrupt_source` part for them. Many thanks.

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.hpp
```diff
@@ -204,6 +205,8 @@ class EXPORT VoodooI2CDeviceNub : public IOService {
     bool has_apic_interrupts {false};
     bool has_gpio_interrupts {false};
     bool use_10bit_addressing {false};
+    bool force_polling { false };
+    IOPCIDevice *pci_device { nullptr };
```
zhen-zen, line 209: Maybe we can just make them local variables since they are not used elsewhere.

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
```diff
@@ -193,17 +201,21 @@ IOReturn VoodooI2CDeviceNub::getDeviceResources() {
...
         (!validateInterrupt() && getAlternativeInterrupt(&crs_parser) == kIOReturnSuccess)) {
         setProperty("gpioPin", crs_parser.gpio_interrupts.pin_number, 16);
         setProperty("gpioIRQ", crs_parser.gpio_interrupts.irq_type, 16);
 
         has_gpio_interrupts = true;
         gpio_pin = crs_parser.gpio_interrupts.pin_number;
         gpio_irq = crs_parser.gpio_interrupts.irq_type;
+        }
+    } else {
+        IOLog("%s::%s Forced polling mode, skipping APIC/GPIO interrupts", controller_name, getName());
```
zhen-zen, line 215: How about moving the prompt above and reduce one more parentheses?

zhen-zen: And missing a `\n` here, I also missed one two lines below lol.

black-dragon74: > And missing a `\n` here, I also missed one two lines below lol.

Is `\n` really required? I've never used it with `IOLog` nor `os_log` (DriverKit).

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.hpp
black-dragon74: Yeah. Sounds right.

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
gongzhenjeef: It will be printed in separate lines with -v. And with Big Sir, kext injected from bootloader can not output to system log. Then it's necessary to grep keywords from `dmesg`.

zhen-zen: It will be printed in separate lines with -v. And with Big Sir, kext injected from bootloader can not output to system log. Then it's necessary to grep keywords from dmesg.

black-dragon74: It still is printed on separate lines (I checked). I omitted `\n` as the statement looks a bit cleaner. Also, in any of Acidanthera kexts, I don't see them using `\n` either.

zhen-zen: They baked it into the macro: https://github.com/acidanthera/Lilu/blob/master/Lilu/Headers/kern_util.hpp#L79

zhen-zen: And some bad example via `sudo dmesg | grep 'Sandbox'`:
```
[ 8571.368915]: Sandbox: com.apple.Safari(455) deny(1) user-preference-read kCFPreferencesAnyApplication[IGFB][LOG  ][TRANSACTIO] [Flip_Transaction] FB0: Flipping display with wsaa 1 flipped from 3840 x 2160 -> 3840 x 2160 on 4234c000 with 801 options at stride 15360
```

black-dragon74: Alright. Got it.

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
```diff
@@ -201,6 +193,17 @@ IOReturn VoodooI2CDeviceNub::getDeviceResources() {
         return kIOReturnNotFound;
     }
 
+    IOPCIDevice *pci_device { nullptr };
+    bool force_polling { false };
+
+    pci_device = controller->nub->controller->physical_device.pci_device;
```
kprinssu, line 199: Please check if `pci_device` is `NULL` before accessing the pointer.

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
```diff
@@ -201,6 +193,17 @@ IOReturn VoodooI2CDeviceNub::getDeviceResources() {
...
+
+    if (checkKernelArg("-vi2c-force-polling")) {
```
kprinssu, line 201: This if statement can be refactored into:

```
bool force_polling  = (pci_device != nullptr) && checkKernelArg("-vi2c-force-polling");
```

black-dragon74: But then we have to check for `force-polling` property too.

kprinssu: Right, missed the else, we want something like:

```
bool force_polling  =  checkKernelArg("-vi2c-force-polling") || ((pci_device != nullptr) && (pci_device->getProperty("force-polling") != nullptr));

```

black-dragon74: Got it, it's for readability. I'll refactor it. Thanks.

kprinssu (approved):
Looks good. Thanks @black-dragon74.

@coolstar @ben9923  Can you please take a look at this PR?

PR #419, opened 2020-12-16
Store artifacts in zip files

ben9923:
Zipping will make downloading artifacts actually possible :P

Check out this PR's artifacts for example.


kprinssu (approved):
👍

PR #445, opened 2021-02-20
Support secondary button

usr-sse2:
This was meant to be a part of VoodooI2C/VoodooI2CHID#39, just forgot to make a PR in this repo.

### DIFF
--- a/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
+++ b/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
@@ -20,12 +20,12 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
     message.contact_count = event.contact_count;
     memset(message.transducers, 0, VOODOO_INPUT_MAX_TRANSDUCERS * sizeof(VoodooInputTransducer));
     
-    VoodooI2CDigitiserTransducer* transducer = (VoodooI2CDigitiserTransducer*) event.transducers->getObject(0);
+    VoodooI2CDigitiserTransducer* firstTransducer = (VoodooI2CDigitiserTransducer*) event.transducers->getObject(0);
 
-    if (!transducer)
+    if (!firstTransducer)
         return MultitouchReturnBreak;
     
-    if (transducer->type == kDigitiserTransducerStylus)
+    if (firstTransducer->type == kDigitiserTransducerStylus)
         stylus_check = 1;
 
     int valid_touch_count = 0;
@@ -48,7 +48,7 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
             valid_touch_count++;
         }
         inputTransducer->isTransducerActive = transducer->tip_switch.value();
-        inputTransducer->isPhysicalButtonDown = transducer->physical_button.value();
+        inputTransducer->isPhysicalButtonDown = false; // it will be passed either as pressure, or as buttons on the "trackpoint" device
         
         inputTransducer->currentCoordinates.x = transducer->coordinates.x.value();
         inputTransducer->previousCoordinates.x = transducer->coordinates.x.last.value;
@@ -67,16 +67,15 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
 
         // Force Touch emulation
         // The button state is saved in the first transducer
-        if (((VoodooI2CDigitiserTransducer*) event.transducers->getObject(0))->physical_button.value() && isForceClickEnabled()) {
+        if (firstTransducer->physical_button.value() && isForceClickEnabled()) {
             inputTransducer->supportsPressure = true;
-            inputTransducer->isPhysicalButtonDown = 0x0;
             inputTransducer->currentCoordinates.pressure = 0xff;
             inputTransducer->currentCoordinates.width = 10;
         }
     }
     
     // set the thumb to improve 4F pinch and spread gesture and cross-screen dragging
-    if (valid_touch_count >= 4 || transducer->physical_button.value()) {
+    if (valid_touch_count >= 4 || firstTransducer->physical_button.value()) {
         // simple thumb detection: to find the lowest finger touch in the vertical direction.
         UInt32 y_max = 0;
         int thumb_index = 0;
@@ -92,6 +91,14 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
         
     super::messageClient(kIOMessageVoodooInputMessage, voodooInputInstance, &message, sizeof(VoodooInputEvent));
     
+    if (!isForceClickEnabled()) {
+        RelativePointerEvent event;
+        event.dx = 0;
+        event.dy = 0;
+        event.buttons = firstTransducer->physical_button.value();
+        event.timestamp = timestamp;
+        super::messageClient(kIOMessageVoodooTrackpointRelativePointer, voodooInputInstance, &event, sizeof(event));
+    }
     return MultitouchReturnBreak;
 }
 


kprinssu:
Looks good!

kprinssu (approved):

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
```diff
@@ -92,6 +91,14 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
...
     super::messageClient(kIOMessageVoodooInputMessage, voodooInputInstance, &message, sizeof(VoodooInputEvent));
     
+    if (!isForceClickEnabled()) {
+        RelativePointerEvent event;
+        event.dx = 0;
+        event.dy = 0;
+        event.buttons = firstTransducer->physical_button.value();
+        event.timestamp = timestamp;
+        super::messageClient(kIOMessageVoodooTrackpointRelativePointer, voodooInputInstance, &event, sizeof(event));
+    }
```
ben9923, line 101: Do we need to send an event on *each* report rather than just reports with a button involved?

ben9923:
@usr-sse2 Thanks!
I figured changes for VoodooI2C were necessary back then.

@usr-sse2 @kprinssu Shouldn't the `isPhysicalButtonDown` field be completely removed from the VoodooInput Simulator message?
I've been thinking it shouldn't be part of it for a long time, as it's not an MT2 feature.
Seems like VoodooPS2 should still be migrated before it can happen, anyway.

BTW, the button bitmask is parsed by `IOHIPointing`, right?

kprinssu:
@ben9923 It's still defined here, https://github.com/acidanthera/VoodooInput/blob/master/VoodooInput/VoodooInputMultitouch/VoodooInputTransducer.h#L43

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
usr-sse2: It's sufficient to send only on state change (both pressing and unpressing).

usr-sse2:
Button is a field in MT2 input report: https://github.com/acidanthera/VoodooInput/blob/bc7fdb5c0c152dcd8c3ba33b5e67bde00b2c853b/VoodooInput/VoodooInputSimulator/VoodooInputSimulatorDevice.hpp#L69
Is this value used by Apple's MT2 drivers? Is it really boolean or a button bitmask?

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
ben9923: In that case, would you want to track the state changes instead, or just keep sending it on every report?

ben9923:
@usr-sse2 Gotcha.
So this PR intends to pretty much ignore it. Nothing should be broken because of that, right?

I really do wonder what Apple's driver do with that value, and whether it's a bitmask (Each finger should have its own pressure value, what can it be used for then?)
@kprinssu Do you have an idea?

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
```diff
@@ -92,6 +91,14 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
         
     super::messageClient(kIOMessageVoodooInputMessage, voodooInputInstance, &message, sizeof(VoodooInputEvent));
     
+    if (!isForceClickEnabled()) {
```
Goshin, line 94:
```suggestion
    if (firstTransducer->physical_button.value() != firstTransducer->physical_button.last.value && !isForceClickEnabled()) {
```

kprinssu: @Goshin what's your reasoning for the above change?

Goshin: As suggested by @ben9923, this is to track the stage change and send the report only when the button is pressed and released.

usr-sse2: Tested with this change, buttons don't work at all. Don't know why, maybe I'll try to debug it later.

ben9923: `VoodooI2CMultitouchHIDEventDriver::setButtonState` will be called twice in each event as there are multiple buttons, making `.last` value pretty much invalid.

vit9696:
I had a brief conversation with @usr-sse2 regarding this change, and we came with a conclusion that while refactoring all the satellites is possible to support value tracking, this is a breaking change that we should not do right now in a minor release.

I actually request this to be merged for the next release, as apparently the changes on the VoodooInput side made one of the issues more apparent for @lvs1974 https://github.com/acidanthera/VoodooInput/commit/a1e92d7e401ae979be293a90c5cb5a033509ed96.

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
Goshin: > VoodooI2CMultitouchHIDEventDriver::setButtonState will be called twice in each event as there are multiple buttons, making .last value pretty much invalid.

So an easy way (might be a little bit ugly) to fix this would be to update the last value only when the value really changes? Something like:

```diff
class TimeTrackedValue {
public:
    ValueWithTime current;
    ValueWithTime last;
...
     void update(UInt16 value, AbsoluteTime timestamp) {
-        last = current;
+        if (current.value != value) {
+            last = current;
+        }
         current.value = value;
         current.timestamp = timestamp;
     }
...
};
```

otherwise, the other ways I can come up with are to refactor `::setButtonState()` to accept two button state values at a time, or to bypass it and manually manipulate the value of `transducer->physical_button`, which is ugly.

ben9923: @Goshin Refactoring (or rather adding `::setButtonMultiState`/whatever) seemed like the proper solution to me. Then I also figured we might have a problem when input is given in multiple digitizer reports.
In such cases state would be updated multiple times anyway (`::handleDigitizerReport` called for each report). Maybe parsing button elements should only be done in first/last report (I wonder how it works in Linux).

Am I right here? CC @kprinssu @usr-sse2

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
```diff
@@ -67,16 +67,15 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
 
         // Force Touch emulation
         // The button state is saved in the first transducer
-        if (((VoodooI2CDigitiserTransducer*) event.transducers->getObject(0))->physical_button.value() && isForceClickEnabled()) {
+        if (firstTransducer->physical_button.value() && isForceClickEnabled()) {
```
ben9923, line 70: Should force touch emulation happen if just the secondary button is down?

ben9923:
@vit9696 Thanks for the input.
Proper tracking is probably only needed for VoodooI2CHID, but as I replied earlier it might be a bigger change than we thought, indeed making it something for a further release.

I do wish to at least have a proper answer for this comment I've made. Not sure if it's really the desired functionality.
```
Also, this line (either with/without last value check) mean secondary button will only work if force touch is disabled in SysPrefs, right?
Do we really want it supported just when it's disabled?
```
I believe that with those changes when force touch is enabled, buttons (actually just primary) will only work when there's a finger on the touchpad.

This PR can lead to a lot of user issues, so I do think it should be closely reviewed even if it doesn't have many changes.


I also have a question - shouldn't `isPhysicalButtonDown` field of VoodooInput be eliminated once VoodooI2C, VoodooRMI and VoodooPS2 migrate to Trackpoint device?

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
Goshin: > Refactoring (or rather adding ::setButtonMultiState/whatever) seemed like the proper solution to me. Then I also figured we might have a problem when input is given in multiple digitizer reports.
In such cases state would be updated multiple times anyway

@ben9923 I agree. How about keeping a `lastPhysicalButtonValue` var in `VoodooI2CNativeEngine`? Will it make it a lot easier? (and a little bit uglier of course.) The last value will be updated only when sending a touch report to VoodooInput, so it can work around those mentioned problems.

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
usr-sse2: Force touch emulation shouldn't be used on touchpads with buttons. It's for touchpads with clickable surface.

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
ben9923: @Goshin Thanks for the confirmation :)
I think keeping the last value should rather be done on the VoodooI2CHID side, as the reported value to the native engine already contains a 'last' value, making it really ambiguous.
IMO another `DigitiserTransducerButtonState` for the button should be a part of the `digitiser`, it will then be properly updated once creating the `VoodooI2CMultitouchEvent` and put into the first transducer.

Also, ideally the button shouldn't be a part of the first transducer but rather a separate filed of the report.
Not sure why it is like that now.

1Revenger1:
> I also have a question - shouldn't `isPhysicalButtonDown` field of VoodooInput be eliminated once VoodooI2C, VoodooRMI and VoodooPS2 migrate to Trackpoint device?

Do we know if macOS does anything with the button field?
VoodooRMI currently does use the trackpoint device for all the buttons other than the clickpad button. I don't see any reason to remove that field since clicking down on the clickpad is analogous to clicking down on a magic trackpad.

usr-sse2:
I suppose that this field is left from the first-generation Magic Trackpad that didn't have force. But it needs to be checked.

theroadw:
I have been testing this PR for the last week on my ZBook G5 17 ALPS trackpad and everything works perfectly.
(Force Touch disabled)

chilledHamza:
Tested this on two devices:-
SYNA2B46 - Lenovo Rescuer Y7000 (Chinese variant of Legion Y530) : Both Buttons work properly
ELAN061B - Lenovo Legion Y530 : primary working as secondary (Left button -> Right button), secondary works as tertiary (Right button -> Middle button)

"Primary Button Element" and "Secondary Button Element" are missing in IOReg under VoodooI2CPrecisionTouchpadHIDEventDriver->Degitizer ([IORegDump](https://github.com/VoodooI2C/VoodooI2C/files/6370192/Hamza.s.MacBook.Pro.zip))

https://github.com/usr-sse2/VoodooI2CHID/blob/d09a771c074483cba48ad769ae701b6649ad5830/VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp#L756-L757

That makes me thing that following piece of code (from [VoodooI2C/VoodooI2CHID/pull/39](https://github.com/VoodooI2C/VoodooI2CHID/pull/39)) is never executed, and primaryButton/secondaryButton is possibly`nullptr` when ::setDigitizerProperties() is called
https://github.com/usr-sse2/VoodooI2CHID/blob/d09a771c074483cba48ad769ae701b6649ad5830/VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp#L583-L600

I don't have SYNA2B46 anymore to test and compare IOReg, so I might be overthinking here.

https://github.com/usr-sse2/VoodooI2CHID/blob/d09a771c074483cba48ad769ae701b6649ad5830/VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp#L265
As a temporary workaround  I'm using :  `setButtonState(&transducer->physical_button, (usage - 2), value, timestamp);`
Fixes button issue for ELAN061B, but will technically break button functionality for other devices.
Just thought I should share my finding with you guys, so that we come up with some proper solution.

lulu-gh:
Hello,

I'm sorry to report that this branch doesn't work on Dell ALPS U1 touchpad, which supports the PTP protocol. After using this branch, I still got only a responsive left button while there are active inputs, and the middle and right buttons are still not working. Please tell me that if there is anything I could do to report bugs. Thanks!

ben9923:
@usr-sse2 As we plan on making a release soon, do you think this is ready for merge?
Is sending the button report on every touch efficient enough, or it would impact performance/energy usage in a noticeable way?

Sorry it took a long while.

usr-sse2:
@ben9923 I use it since making this PR and haven't noticed any impact on the performance or energy consumption.

ben9923:
@usr-sse2 In that case, we can go ahead and merge :)

CC @kprinssu

ben9923 (approved):
There are  some optimizations & improvements I'd like to see in this code, as discussed in the review threads, but it looks fine for now.

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
kprinssu: We can address this in a follow-up PR. It's good as-is.

kprinssu (approved):

PR #473, opened 2021-09-10
Remove overriding bus timing defaults

williambj1:
Hi devs!

This PR allows accepting a single set of bus `*CNT` values (`hcnt`, `lcnt`, and `sda hold`) for a certain bus frequency (Standard Mode and Fast Mode) and leaves the missing sets as default. Which will actually reflect the log that says "using defaults where necessary".

https://github.com/VoodooI2C/VoodooI2C/blob/9ddc72ab7e3ff1007a534dc425977fd7d7d298a3/VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CControllerDriver.cpp#L468

Previously if one set of the `*CNT` values for a certain frequency is missing, all of the values will be set back to default. This is problematic on platforms where the OEM only provides a set of customized values for the desired frequency and leaves others unset.

Here is an example from an ASUS laptop (Zephyrus S GX531GS):

```asl
// Line 60904
Scope (_SB.PCI0.I2C1)
{
    // Only FMCN is provided
    Method (FMCN, 0, NotSerialized)
    {
        Name (PKG, Package (0x03)
        {
            0x0101, 
            0x012C, 
            0x62
        })
        Return (PKG)
    }
    Device (ETPD)
    {
        Name (SBFB, ResourceTemplate ()
        {                                                // 0x61A80 is 400KHz, Fast Mode
            I2cSerialBusV2 (0x004C, ControllerInitiated, 0x00061A80,
                AddressingMode7Bit, "\\_SB.PCI0.I2C1",
                0x00, ResourceConsumer, _Y34, Exclusive,
                )
        })
        Name (SBFI, ResourceTemplate ()
        {
            Interrupt (ResourceConsumer, Level, ActiveHigh, Exclusive, ,, )
            {
                0x0000005F,
            }
        })
        // ...
```

Original DSDT for reference:
[DSDT.aml.zip](https://github.com/VoodooI2C/VoodooI2C/files/7140824/DSDT.aml.zip)

---

BTW, I realized that the controller code is ported from the Linux DesignWare I2C driver and there is a way to read the tx and rx fifo depth. Currently the tx and rx fifo depth is hardcoded, is there any particular reason for that?

https://github.com/torvalds/linux/blob/bf9f243f23e6623f310ba03fbb14e10ec3a61290/drivers/i2c/busses/i2c-designware-common.c#L572-L598

Thanks and have a nice day.

### DIFF
--- a/VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CControllerDriver.cpp
+++ b/VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CControllerDriver.cpp
@@ -466,11 +466,6 @@ bool VoodooI2CControllerDriver::start(IOService* provider) {
 
     if (getBusConfig() != kIOReturnSuccess) {
         IOLog("%s::%s Warning: Error getting bus config, using defaults where necessary\n", getName(), bus_device.name);
-        bus_device.acpi_config.ss_hcnt = 0x01b0;
-        bus_device.acpi_config.fs_hcnt = 0x48;
-        bus_device.acpi_config.ss_lcnt = 0x01fb;
-        bus_device.acpi_config.fs_lcnt = 0xa0;
-        bus_device.acpi_config.sda_hold = 0x9;
     } else {
         IOLog("%s::%s Got bus configuration values\n", getName(), bus_device.name);
     }


ben9923:
Thanks for this, it's indeed a welcome change.
I do wonder if the `VoodooI2CControllerBusConfig` defaults are enough so no setting should be done here on failure whatsoever.

BTW, I plan to support per-platform defaults in the near future, as the current defaults are Haswell era (`Defined hsw_config in Linux`)...
We've seen they don't usually fit CFL/CNL+ hardware, requiring custom SSDTs with SSCN/FMCN.
Correctly implemented for over 5 years in the Linux LPSS code (Used for SKL+) :P

Regarding FIFO size, I don't know. Could legacy hardware (Haswell) used hardocded values in the past?
You can look at the Linux changelog, I'll do so too later.

williambj1:
@ben9923 Thank you very much for the reply. Per-platform defaults will certainly be a great improvement.

> I do wonder if the VoodooI2CControllerBusConfig defaults are enough so no setting should be done here on failure whatsoever.

I'm not sure if there are cases where the buffer returned from FMCN/SSCN is incomplete/invalid. If we don't care about it (I don't think linux does either), then it's fine to remove the assignments.

I haven't looked at the git history regarding the FIFO size, will probably do so tomorrow.

BTW, the DesignWare Driver has a way to calculate/approximate the missing `*CNT` values. But I did not understand the `ic_clk`.

- https://github.com/torvalds/linux/blob/6880fa6c56601bb8ed59df6c30fd390cc5f6dd8f/drivers/i2c/busses/i2c-designware-common.c#L335-L387
- https://github.com/torvalds/linux/blob/6880fa6c56601bb8ed59df6c30fd390cc5f6dd8f/drivers/i2c/busses/i2c-designware-master.c#L40-L165

ben9923:
> I'm not sure if there are cases where the buffer returned from FMCN/SSCN is incomplete/invalid. If we don't care about it (I don't think linux does either), then it's fine to remove the assignments.
> 
Looking at the `getACPIParams` implementation, I think it's safe to use the defaults and not re-assign at all.

> I haven't looked at the git history regarding the FIFO size, will probably do so tomorrow.
> 
It appears to originate from the very first Lynx Point dw commit: https://github.com/torvalds/linux/commit/b61b14154b19e1ef1da9c1e283f0cf93470e0c70
It was removed 1.5 years ago (So relatively new...): https://github.com/torvalds/linux/commit/3f35064a7cfef4ed8d25cdb16da0abfbbd525f63
Currently implemented here:
https://github.com/torvalds/linux/blob/6880fa6c56601bb8ed59df6c30fd390cc5f6dd8f/drivers/i2c/busses/i2c-designware-common.c#L572-L598

I have no idea if any intel PCH has a FIFO size that's higher than 32, but it's simple enough to add for the chance it'll optimize some platforms. I'll try looking those values up on 400 series, etc.


> BTW, the DesignWare Driver has a way to calculate/approximate the missing `*CNT` values. But I did not understand the `ic_clk`.
> 
> * https://github.com/torvalds/linux/blob/6880fa6c56601bb8ed59df6c30fd390cc5f6dd8f/drivers/i2c/busses/i2c-designware-common.c#L335-L387
> * https://github.com/torvalds/linux/blob/6880fa6c56601bb8ed59df6c30fd390cc5f6dd8f/drivers/i2c/busses/i2c-designware-master.c#L40-L165
>
Yeah, that's what I was referring to. `ic_clk` is provided in the LPSS driver for SKL+. Haswell doesn't need it as SCL/SDA values are hardcoded (See `i2c_dw_clk_rate` docs).

I can do both :)

ben9923:
@williambj1 Accidentally messed up the remote while trying to rebase :/
Could you re-push the changes rebased with upstream? I'll be able to re-open once you do so.

Sorry!
(I'll remove that silly requirement for a rebase/merge of master into the source branch prior to merge)

williambj1:
@ben9923 Done :)

GitHub doesn't allow me to reopen the PR, please let me know if a new PR should be created. Thanks!

ben9923 (approved):
Thanks!

kprinssu (approved):
Looks good to me.

zhen-zen:
Specific to this issue, maybe it's better to check the return of `validateObject` first? We don't need to proceed nor report error in case of `kIOReturnNotFound`.

ben9923:
@zhen-zen I believe we should treat it as an error until proper default bus config exists for SKL+, as currently default aren't good.

Once per-platform defaults feature is merged - You're right, it should not be treated as an error.

PR #480, opened 2021-09-19
Add Missing Trailing Newlines in Logs

ben9923:
Those logs were printed together with other random system logs in the same lines :/

### DIFF
--- a/VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CController.cpp
+++ b/VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CController.cpp
@@ -49,7 +49,7 @@ IOReturn VoodooI2CController::unmapMemory() {
 VoodooI2CController* VoodooI2CController::probe(IOService* provider, SInt32* score) {
     if (!super::probe(provider, score)) {
         if (debug_logging)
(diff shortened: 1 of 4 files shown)


kprinssu (approved):
Looks good. Do we need to update the other IOLogs calls across the project as well?

1Revenger1:
All IOLog statements should have new lines on them
I ended up creating a [macro for it](https://github.com/VoodooSMBus/VoodooRMI/blob/90d313d29eca1e2b335b59d057100d370c61f489/VoodooRMI/Utility/Logging.h#L15) in VRMI.
May also be able to put in getName() in there as well if you wanted, doesn't work when outside an IOService class though, which is the issue I ran into.

ben9923:
@kprinssu There are a couple of similar cases in the other kexts, I can push those too.

@1Revenger1 Isn't `log` broken (i.e. not showing all kext logs)?
Otherwaise why we ask users for `dmesg` output by stretching `msgbuf`?

1Revenger1:
> @1Revenger1 Isn't `log` broken (i.e. not showing all kext logs)?
> Otherwaise why we ask users for `dmesg` output by stretching `msgbuf`?

Useful for streaming data in real time. Although that's broke half the time as well lol.

PR #483, opened 2021-10-13
VoodooI2CControllerNub: Release object when values not found in ACPI

stevezz3:
This commit could prevent potential memory leaks when FMCN/SSCN not found in ACPI table.
Big thanks to @williambj1


kprinssu (approved):
This looks good to me. @ben9923 Please take look as well.

Issue #487, opened 2021-12-07
Almost silent failure mode of GPIO pinning resulting in polling while IORegistry suggests success


gvkt:
There is an almost silent mode of pinning failure which makes the kexts resort to polling even when the GPIO pinning seems to have succeeded (as verified via just looking at the IORegistry after the change). This is only evident by looking at the elusive logs that most people won't find and they will think GPIO pinning has succeeded while it is doing polling. IORegistry will list the GPIO pin and interrupt even though polling is in use.

Since many incorrect pin numbers can cause this situation, people may think they have switched to interrupt based reports while it is doing polling underneath.

This is when an incorrect pin number leads to VoodooI2CHIDDevice::handleStart not finding an interrupt source for the specified pin and switches to polling. This failure is not reflected in the IORegistry properties set for the pin number and interrupt and looks as if GPIO pinning has succeeded.

It might be useful to set an explicit flag in the device properties for polling being true as default which is set to false only on successful interrupt set up (or vice versa) so a confirmation can be done via looking at IORegistry as to whether polling or interrupt is the actual mode in use.

ben9923:
Sounds good to me, especially when logs seem to be broken (Requiring users to look at `dmesg`, only possible by increasing `msgbuf` - manually or via DebugEnhancer).

shiecldk:
@gvkt I found sqlsec mentions how he derive his gpio pin here: https://www.sqlsec.com/2021/12/screenpad.html
You may want to use Google Translate to read the article.

kprinssu:
This has been addressed by @1Revenger1(https://github.com/VoodooI2C/VoodooI2C/pull/493) and there is now an IO Registry to show if you are using interrupts or polling.

PR #493, opened 2022-02-01
Display interrupt type in IOReg

1Revenger1:
It's a little hard sometimes to figure out whether VoodooI2C is polling or using APIC/GPIO interrupts without looking at logs. This provides a quick way to figure out what mode is being used. Solves #487 

Sorry for the messy commits, was from me fooling around with VoodooI2CHID at the same time.

### DIFF
--- a/VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
+++ b/VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
@@ -19,6 +19,8 @@
 OSDefineMetaClassAndStructors(VoodooI2CDeviceNub, IOService);
 
 bool VoodooI2CDeviceNub::attach(IOService* provider, IOService* child) {
+    char const *interruptMode = "Polling";
+    
     if (!super::attach(provider))
         return false;
 
@@ -44,6 +46,7 @@ bool VoodooI2CDeviceNub::attach(IOService* provider, IOService* child) {
     }
 
     if (has_gpio_interrupts) {
+        interruptMode = "GPIO";
         gpio_controller = getGPIOController();
 
         if (!gpio_controller) {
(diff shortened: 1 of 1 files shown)


narcyzzo (approved):
Hi, I tested those changes, really very usefull.

VoodooI2C/VoodooI2C/VoodooI2CDevice/VoodooI2CDeviceNub.cpp
```diff
@@ -19,6 +19,8 @@
 OSDefineMetaClassAndStructors(VoodooI2CDeviceNub, IOService);
 
 bool VoodooI2CDeviceNub::attach(IOService* provider, IOService* child) {
+    char const *interruptMode = "Polling";
```
ben9923, line 22: Can you switch to the more-common `const char *` prefix, to keep it the same as the rest of the code?

ben9923, line 22: It might be more explicit/readable to assign a value within the conditions, something like this:

```C++
if (has_gpio_interrupts) {
    ...

     interruptMode = "GPIO";
} else if (has_apic_interrupts) {
    interruptMode = "APIC";
} else {
    interruptMode = "Polling";
}
```

1Revenger1: Just pushed, this closer to what you were looking for?

ben9923: I think the linter wouldn't like it not initialized as `nullptr`, but yeah, thanks :)

1Revenger1: The linter didn't seem to care, but I solved a couple other lint things. Will just add nullptr anyways.

ben9923 (approved):

kprinssu (approved):

kprinssu:
@1Revenger1 Can you make a small change and push? Looks like our CI pipeline is broken and the push will cause it to re-build.

1Revenger1:
Does circleci not like that I'm working on a branch other then master? I don't have any experience with circleci, mostly work with github actions.

Issue #499, opened 2022-03-16
In Recovery clicking with separate physical button requires a finger to be on the touchpad


usr-sse2:
If there are no fingers on the touchpad, clicks are ignored.

kprinssu:
@usr-sse2 Does https://github.com/VoodooI2C/VoodooI2C/commit/9ab98319b6411c4a4822cc12f840df85cc684bfc address this? It contains your changes from PR#503.

kprinssu:
Closing as this should be fixed by [`9ab98319b6411c4a4822cc12f840df85cc684bfc`](https://github.com/VoodooI2C/VoodooI2C/commit/9ab98319b6411c4a4822cc12f840df85cc684bfc). Please comment if it is not.

PR #503, opened 2022-04-19
Fix multiple issues with clicking

### LINKED ISSUE #496: Right click using click on bottom right doesn't work anymore
**Describe the bug**
Right click doesn't work anymore on latest VoodooI2C version

**Did you read the common errors documentation?**
yes

**Did you read the troubleshooting documentation?**
yes

**Have you searched the issue on Github, Gitter, or Google?**
yes

**System Environment**
Please answer:
1. What is your operating system?  Big Sur 11.6.4
2. What version of the drivers (including drivers) are you using? 2.7 latest
3. What is your hardware configuration? Intel i5 9th gen - 16GB RAM - 512GB SSD

I've upgraded today to the latest VodooI2C version but after reboot i discovered that right click using bottom right setting in trackpad doesn't work anymore. If i chose right click using tap with two fingers it works as expected. Before the update was working fine in both ways. Gesture are working ok.
Basically even if i click on the bottom right of my touchpad is always considered a left click

Mattia

usr-sse2:
1. Always pass buttons as pressure on single-button touchpads (fix #496)
2. Always disable Force Touch and pass buttons to buttons device on touchpads with two physical buttons (fixes inability to click without finger on the touchpad (#499) and inability to use the right button with default settings)
3. Disable Force Touch on all touchpads if tap to click is disabled (fixes inability to make a non-force click with default settings, especially in Recovery where there is no System Preferences application)

### DIFF
--- a/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
+++ b/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
@@ -48,7 +48,7 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
             valid_touch_count++;
         }
         inputTransducer->isTransducerActive = transducer->tip_switch.value();
-        inputTransducer->isPhysicalButtonDown = false; // it will be passed either as pressure, or as buttons on the "trackpoint" device
+        inputTransducer->isPhysicalButtonDown = !transducer->has_secondary_button && transducer->physical_button.value(); // if it has secondary button, then it will be passed as buttons on the "trackpoint" device
         
         inputTransducer->currentCoordinates.x = transducer->coordinates.x.value();
         inputTransducer->previousCoordinates.x = transducer->coordinates.x.last.value;
@@ -67,8 +67,10 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
 
         // Force Touch emulation
         // The button state is saved in the first transducer
-        if (firstTransducer->physical_button.value() && isForceClickEnabled()) {
+        // Disable Force Touch emulation for two-button trackpads
+        if (!firstTransducer->has_secondary_button && firstTransducer->physical_button.value() && isForceClickEnabled()) {
             inputTransducer->supportsPressure = true;
+            inputTransducer->isPhysicalButtonDown = false;
             inputTransducer->currentCoordinates.pressure = 0xff;
             inputTransducer->currentCoordinates.width = 10;
         }
@@ -91,7 +93,7 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
         
     super::messageClient(kIOMessageVoodooInputMessage, voodooInputInstance, &message, sizeof(VoodooInputEvent));
     
-    if (!isForceClickEnabled()) {
+    if (firstTransducer->has_secondary_button) {
         RelativePointerEvent event;
         event.dx = 0;
         event.dy = 0;
@@ -111,7 +113,7 @@ bool VoodooI2CNativeEngine::start(IOService* provider) {
     if (!parentProvider) {
         return false;
     }
-
+    
     voodooInputInstance = NULL;
     
     setProperty(VOODOO_INPUT_LOGICAL_MAX_X_KEY, parentProvider->logical_max_x, 32);
@@ -137,7 +139,7 @@ bool VoodooI2CNativeEngine::isForceClickEnabled() {
         lastForceClickPropertyUpdateTime = now_abs;
         return lastIsForceClickEnabled;
     }
-
+    
     // Blocking reading here takes about 10 microseconds
     OSDictionary* dict = OSDynamicCast(OSDictionary, this->getProperty("MultitouchPreferences", gIOServicePlane, kIORegistryIterateRecursively));
     if (!dict) {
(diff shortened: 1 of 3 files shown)


Multitouch Support/Native/VoodooI2CNativeEngine.cpp
```diff
@@ -137,14 +139,32 @@ bool VoodooI2CNativeEngine::isForceClickEnabled() {
...
     OSDictionary* dict = OSDynamicCast(OSDictionary, this->getProperty("MultitouchPreferences", gIOServicePlane, kIORegistryIterateRecursively));
     if (!dict) {
         return lastIsForceClickEnabled;
     }
     if (OSCollectionIterator* i = OSCollectionIterator::withCollection(dict)) {
         while (OSSymbol* key = OSDynamicCast(OSSymbol, i->getNextObject())) {
+            // If tap to click (property "Clicking") is disabled,
+            // then we should forbid force clicking,
+            // because it makes impossible to perform a non-force click
+            if (key->isEqualTo("Clicking")) {
```
kprinssu, line 153: Nitpick: There is no guarantee that the property will not change in the future. We can refactor this method to handle any such situations in the future.

Stylistic: Can we perhaps pull the string `"Clicking"` into a `#define` constant?

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
```diff
@@ -137,14 +139,32 @@ bool VoodooI2CNativeEngine::isForceClickEnabled() {
...
+                OSBoolean *boolValue;
+                OSNumber *numValue;
+                if ((boolValue = OSDynamicCast(OSBoolean, dict->getObject(key))) != NULL) {
```
kprinssu, line 156: Stylistic: This is a bit hard to read, can we perhaps please refactor the conidition to the following:

```
boolValue = OSDynamicCast(OSBoolean, dict->getObject(key);
if (boolValue != NULL && !boolValue->getValue()) { ....

```

This reduces the extra if statement as well.

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
```diff
@@ -137,14 +139,32 @@ bool VoodooI2CNativeEngine::isForceClickEnabled() {
...
+                    if (!boolValue->getValue()) {
+                        lastIsForceClickEnabled = FALSE;
+                        break;
+                    }
+                }
+                // is it needed? in default properties it's number, not boolean
+                else if ((numValue = OSDynamicCast(OSNumber, dict->getObject(key))) != NULL && !numValue->unsigned64BitValue()) {
```
kprinssu, line 163: Stylistic: This is also rather hard to read, can we please pull the assignment out of the if condition?

kprinssu (changes requested):
Thanks @usr-sse2! There are some minor stylistic comments, it would be great to add them in :)!

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
usr-sse2: No, it doesn't reduce the extra `if` statement. The if should be nested, because `else` should be executed only if cast failed (it's not a boolean), but not if `getValue()` returned true.

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
usr-sse2: I can pull it out of condition, but it will increase the nesting level, because the cast should be executed only if the previous cast to boolean did not succeed. (In fact, I have originally written it that way).

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
ben9923, line 156: Please convert `NULL -> nullptr` (not just here)

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
```diff
@@ -137,14 +139,32 @@ bool VoodooI2CNativeEngine::isForceClickEnabled() {
...
         while (OSSymbol* key = OSDynamicCast(OSSymbol, i->getNextObject())) {
+            // If tap to click (property "Clicking") is disabled,
+            // then we should forbid force clicking,
+            // because it makes impossible to perform a non-force click
+            if (key->isEqualTo("Clicking")) {
+                OSBoolean *boolValue;
+                OSNumber *numValue;
+                if ((boolValue = OSDynamicCast(OSBoolean, dict->getObject(key))) != NULL) {
+                    if (!boolValue->getValue()) {
+                        lastIsForceClickEnabled = FALSE;
```
ben9923, line 158: Please convert `FALSE -> false` (not just here)

Multitouch Support/VoodooI2CDigitiserTransducer.hpp
```diff
@@ -90,6 +90,7 @@ class EXPORT VoodooI2CDigitiserTransducer : public OSObject {
     
 public:
     DigitiserTransducerButtonState physical_button;
+    bool has_secondary_button = false;
```
ben9923, line 93: I'll create a thread here because I can't do it on the satellite diff:
We can't just change the VoodooI2CHID commit, it's not part of its `master` branch :P

BTW, do we really need to set `has_secondary_button` on every report? It will never change.

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
```diff
@@ -137,14 +139,32 @@ bool VoodooI2CNativeEngine::isForceClickEnabled() {
...
+                        break;
+                    }
+                }
+                // is it needed? in default properties it's number, not boolean
```
ben9923, line 162: Really Apple? :(

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
ben9923: As other code in this method isn't using consts for keys IDK if you really need to do it now...

ben9923 (changes requested):
Thanks for your hard work Michael!
I believe it will resolve some annoying issues users experience...

With those stytlistic comments there's the satellite change that should probably happen in another PR (Or two actually - one for HID, then one to change it here).

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
usr-sse2: Default preferences:
Знімок екрана 2022-05-13 о 12 39 02

Current preferences:
Знімок екрана 2022-05-13 о 12 39 56

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
usr-sse2: Done

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
usr-sse2: Done

Multitouch Support/Native/VoodooI2CNativeEngine.cpp
Goshin: If that is the case, can we safely presume that `lastIsForceClickEnabled` should be `false` if the property is not a boolean value? so that we can simplify this a bit as there is no need to read out the number value.

kprinssu (approved):

kprinssu (approved):

kprinssu (approved):

kprinssu:
Addressed with https://github.com/VoodooI2C/VoodooI2C/pull/514

PR #523, opened 2023-01-02
Bump to version 2.7.1

kprinssu:
This will incorporate proper changes from @usr-sse2 to add support for physical left and right buttons. As well as fixing panics on shutdown for USB based devices.


ben9923 (approved):
Cool!

PR #547, opened 2024-04-17
Add confidence bit for palm rejection

1Revenger1:
I was parsing the MT2 output for some other projects and came across a new finger ID. This seems to only be sent when the touchpad detects a palm. This seems to map well to the confidence bit sent by Precision/HID touchpads, and allow for rejecting contacts based off of size relatively easily.

This does break every satellite since the Transducer class is modified, so some caution is needed before trying to merge this in.

Companion ELAN implementation:
https://github.com/1Revenger1/VoodooI2CELAN/pull/1

Companion HID implementation:
https://github.com/1Revenger1/VoodooI2CHID/pull/1

### DIFF
--- a/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
+++ b/Multitouch Support/Native/VoodooI2CNativeEngine.cpp
@@ -30,6 +30,7 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
 
     int valid_touch_count = 0;
 
+    IOLog("I2C Contact Count: %d\n", event.contact_count);
     for (int i = 0; i < event.contact_count; i++) {
         VoodooI2CDigitiserTransducer* transducer = (VoodooI2CDigitiserTransducer*) event.transducers->getObject(i+stylus_check);
         VoodooInputTransducer* inputTransducer = &message.transducers[i];
@@ -47,6 +48,11 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
         if (inputTransducer->isValid) {
             valid_touch_count++;
         }
+        
+        if (!transducer->confidence.value()) {
+            inputTransducer->fingerType = static_cast<MT2FingerType>(6);
+        }
+        
         inputTransducer->isTransducerActive = transducer->tip_switch.value();
         inputTransducer->isPhysicalButtonDown = !transducer->has_secondary_button && transducer->physical_button.value(); // if it has secondary button, then it will be passed as buttons on the "trackpoint" device
         
@@ -74,6 +80,8 @@ MultitouchReturn VoodooI2CNativeEngine::handleInterruptReport(VoodooI2CMultitouc
             inputTransducer->currentCoordinates.pressure = 0xff;
             inputTransducer->currentCoordinates.width = 10;
         }
+        
+        IOLog("I2C [%i] - Valid: %d - Confidence: %d - Tip Switch: %d\n", i, transducer->is_valid, transducer->confidence.value(), transducer->tip_switch.value());
     }
     
     // set the thumb to improve 4F pinch and spread gesture and cross-screen dragging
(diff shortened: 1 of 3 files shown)


kprinssu:
I think we can preserve backwards compatibility by moving the field to the end of the struct.

This should ensure that memory wise, the first N bytes will hold the existing transducer data.

1Revenger1:
Ah that would solve it. We should probably have it default to a value of 1 as well so that everything isn't read in as a palm - what is the best way to do that?

kprinssu:
> Ah that would solve it. We should probably have it default to a value of 1 as well so that everything isn't read in as a palm - what is the best way to do that?
> 
> 
> 
> Edit: oh would there be any worries about out of bound reads if we did just add it to the end?

That would indeed cause out of bound reads. Another option is to make the change and tell people to update their satellite lets.

IMO, this has been the case for a while and I don't see any issue with us changing the class.

1Revenger1:
> That would indeed cause out of bound reads. Another option is to make the change and tell people to update their satellite lets.
> 
> IMO, this has been the case for a while and I don't see any issue with us changing the class.

Looking around, I think that is a safe assumption that can be made. The satellites compiled as part of VoodooI2C are the only ones to use this class. VoodooRMI and AlpsHID do not use the VoodooI2C transducer, instead interacting directly with VoodooInput.

I think this PR is good to go as well as the Elans PR.
The HID PR should be good to go for precision touchpads, although I want to try using width/height to reject contacts on touchscreens before I undraft it.

kprinssu:
Thanks @1Revenger1. I'll merge these in when I get the chance to fix the pipelines.

1Revenger1:
No worries. If anyone wants to give these a shot, I've got a download here with VoodooI2C, VoodooI2CHID, and VoodooI2CELAN.
[Archive.zip](https://github.com/VoodooI2C/VoodooI2C/files/15102594/Archive.zip)

Goshin:
Hi, there.

Since it is related to MT2 simulating, would it be better to add the palm type in `MT2FingerType` definition of VoodooInput? Besides better readability, other drivers using VoodooInput could also benefit from this type for palm rejection.

https://github.com/acidanthera/VoodooInput/blob/8e36453623155df94b930fb24c8d22e10f73dd1b/VoodooInput/VoodooInputMultitouch/VoodooInputTransducer.h#L13

1Revenger1:
Probably should yeah. I can add it later tonight.
I wonder how much of the palm rejection stuff we could move into VoodooInput. We should really be forwarding everything to macOS even when typing, so I don't see a reason why every driver should need to do keyboard quiet time or whatever by itself.

kprinssu:
> Probably should yeah. I can add it later tonight. I wonder how much of the palm rejection stuff we could move into VoodooInput. We should really be forwarding everything to macOS even when typing, so I don't see a reason why every driver should need to do keyboard quiet time or whatever by itself.

It was a hack from when Rehanman was maintaining VoodooPS2 (so about 6-7 years back).

1Revenger1:
> It was a hack from when Rehanman was maintaining VoodooPS2 (so about 6-7 years back). IPC in kernel-space is tough so abusing the quiet time field to receive notifications from the keyboard driver made a lot of sense.

Yeah, and we haven't really had a reason to forward invalid things to macOS before this since macOS would have treated it as valid data. I'm not faulting anyone here, palm rejection is just a pain in general haha.

1Revenger1:
I opened the VoodooI2CHID PR and merged in the new finger type into VoodooInput, so everything should be good to go.

LeeBinder:
> No worries. If anyone wants to give these a shot, I've got a download here with VoodooI2C, VoodooI2CHID, and VoodooI2CELAN. [Archive.zip](https://github.com/VoodooI2C/VoodooI2C/files/15102594/Archive.zip)

@1Revenger1 thank you - working fine in Monterey (Sonoma not tested yet) in my Asus Vivobook S15 with an ELAN1200 (VID 0x4f3 | PID 0x303e)

PR #574, opened 2025-11-12
Rework default bus timings

averycblack:
NOTE: I would like testers on this, especially for AMD and older systems. Binaries can be found in github actions.
This is staying as a draft for now until more tests are done. @kprinssu 

Finally bit the bullet and decided to rework how bus timings are done.
Linux has several different ways to add config values but I decided to throw them all into IOKit personalities/properties here.

As part of IOKit Personalities, several properties added to `VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CControllerConstants.hpp` may be used to define SDA hold times, hold counts, etc. A new requirement for any new I2C controller to be added is that I2CClkRate must be set - as otherwise none of the other values may be derived (1).

VI2C should now:
1) Grab default values from the controllers (if any) and clock
2) Grab SSCN and FMCN from ACPI
3) If any values are not filled in, calculate some sane defaults.

My Linux references for these values:
- Intel Skylake+:
  - /drivers/mfd/intel-plss-pci.c
- Intel Broadwell-:
  - /drivers/i2c/busses/i2c-designware-pcidrv.c
  - /drivers/acpi/x86/plss.c
- AMD:
  - /drivers/i2c/busses/i2c-designware-platdrv.c
  - /drivers/acpi/acpi_apd.c
  - /drivers/i2c/busses/i2c-designware-pcidrv.c * Some Navi devices only

(1) Clock is optional in Linux I think, but I figured it was a fair assumption here since all the clocks are relatively easy to find.


VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CControllerDriver.cpp
```diff
@@ -84,44 +107,58 @@ IOReturn VoodooI2CControllerDriver::getBusConfig() {
...
 
-    auto i2c_clk = getClkRateFor(nub->controller->physical_device.name);
+    memset(&bus_device.acpi_config, 0, sizeof(bus_device.acpi_config));
 
-    bool is_sunrise_point = !strncmp(nub->controller->physical_device.name, "INT344B", sizeof("INT344B"))
-            || !strncmp(nub->controller->physical_device.name, "INT345D", sizeof("INT345D"));
+    UInt32 clkKHz = getNumProperty(kI2CPropClkKey) / 1000;
+    if (!clkKHz) {
+        IOLog("%s::%s Failed to find clock!\n", getName(), bus_device.name);
+        return kIOReturnError;
```
averycblack, line 114: Any idea what sort of behavior we would want if we run into this? Returning an error doesn't stop VoodooI2C from loading up. Maybe an assert? If we run into this, someone has added a bad IOKit personality so I'd be fine stopping VI2C or even panicking haha.

averycblack:
Been tested on an Acer Nitro 5 (Coffee lake refresh) and Acer Spin 713 Chromebook (Cometlake)

VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CControllerDriver.cpp
kprinssu: I suggest we handle the return code higher up in the call stack and gracefully not load the driver.

averycblack:
Got some additional tests with a Dell Latitude 5580 (Kabylake/sunrise point, thanks @Corpnewt) and Dell 5591 (Coffelake/cannon lake, thanks theroadw). I believe that covers all of the VoodooI2CPCIController cases.

averycblack:
Tested on an HP Victus 15 with AMD. @kprinssu think that is about as good as we will get for testing.

kprinssu:
Thanks @corpnewt and @averycblack 

I will merge this in sometime in the next couple of days.

PR #246, opened 2020-02-18
Fix Unknown Synopsys component type issue in Comet Lake 2nd

startpenghubingzhou:
This commit has fixed the issue #236 . This work is based on https://www.notion.so/Using-VoodooI2C-on-comet-lake-cpu-e-g-i5-10210u-142930887087445eaa533120455da5dc , which will force D0 in Comet Lake. And I completed it to check the platform to avoid this patch to be applied in other platforms. See code comments for more details.

PR #575, opened 2025-11-23
Use IOPCIFamily helpers to set D0 state for PCIe controllers

averycblack:
Based on @zhen-zen and @startpenghubingzhou PRs. Leaving as draft until #574 is merged since this will need a rebase/merge de-conflict. Removed the Icelake controller class as this logic should work for both ice/comet lake lake and older devices as long as the PCIe power management capability is used.

Tested on my Acer spin 713 (Chromebook)


VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CPCIController.cpp
```diff
@@ -12,10 +12,24 @@
...
+     * Based on PR #246 (credit @startpenghubingzhou and @phu54321)
+     */
+    if (pci_device->extendedFindPCICapability(kIOPCIPowerManagementCapability, &offset)) {
+        IOLog("%s::%s Setting PCI Power state D0\n", getName(), physical_device.name);
+        reg = pci_device->extendedConfigRead32(offset + 4);
+        reg = (reg & ~kPCIPMCSPowerStateMask) | kPCIPMCSPowerStateD0;
+        pci_device->extendedConfigWrite32(offset + 4, reg);
+    }
+    
+    pci_device->enablePCIPowerManagement(kPCIPMCSPowerStateD3);
```
averycblack, line 32: enablePCIPowerManagement does not meaningfully change the power state of the underlying device. It allows or prevents the controller from sleeping when setPowerState is called, but macOS makes the assumption that the device is already at D0 when booting. Because of this assumption, macOS will never touch the PM registers until the laptop is put to sleep as far as I can tell. Allowing D3 power state when the device is sleeping I think should be ok.

kprinssu:
Thanks @averycblack!

startpenghubingzhou:
Great Jobs! @averycblack  As I have mentioned in the #246 , it' s a temporary solution. With this new PR, everything can be more perfectly. Thx for your hard work!

PR #530, opened 2023-03-12
Add AMD support

ChefKissInc:

PR #576, opened 2025-11-23
Simplify memory mapping and only turn on clock gating on legacy Intel

averycblack:
Corrects a change in #530 where clock gating was being applied to AMD platforms.
Looking around, I'm like 90% certain this should only be applied to older Intel platforms which used the [lpss driver](https://github.com/torvalds/linux/blob/master/drivers/acpi/x86/lpss.c#L440) in Linux. Sunpoint+ [does not require this](https://github.com/torvalds/linux/blob/master/drivers/mfd/intel-lpss.c#L359).

Draft until https://github.com/VoodooI2C/VoodooI2C/pull/574 is merged. Be cool to get someone to test this as I don't have a Broadwell or older system.

While in here, I also tried to simplify the mapping where possible. There shouldn't be any reason why memory needs to be mapped repeatedly. Using the IOPCIFamily mapping functions also [allows macOS to move BAR addresses freely](https://developer.apple.com/library/archive/documentation/HardwareDrivers/Conceptual/ThunderboltDevGuide/Basics02/Basics02.html#//apple_ref/doc/uid/TP40011138-CH4-SW13) without changing our virtual address. This has been tested on my Acer Spin 713 chromebook (sleep + hibernate).


averycblack:
Lint failed for some reason, any possibility to rerun it?
I've just now seen this PR here (https://github.com/VoodooI2C/VoodooI2C/pull/258) which I guess explains why we were mapping/unmapping. Is it fair to assume that maybe we were using the wrong function to map for PCIe devices initially? I'm fine with reverting the changes if you don't want to chance it.

Edit: the old mapping function was supposed to respond to BARs moving around too. I'll just redo this PR with only the clock gating changes.

PR #577, opened 2025-11-26
Move AMD quirk handling into VoodooI2CControllerDrivers

averycblack:
Mostly in preparation for fixing the LPSS gating.
Still unsure how I want to handle clock gating but surely making the ACPI class simpler will help. Makes the AMD quirk similar to how defaults are handled now for bus config too.


averycblack:
Tested on an HP Victus 15 (AMD)

kprinssu:
Thanks again for refactoring this bit of the code @averycblack

PR #578, opened 2025-11-26
Apply clock gating only to Broadwell/Lynxpoint I2C controllers

averycblack:
Sunrisepoint+ nor AMD cannot gate their clocks and do not have the LPSS registers that Broadwell/Lynxpoint have.


VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CACPIController.cpp
```diff
@@ -7,10 +7,18 @@
...
 #include "VoodooI2CACPIController.hpp"
+#include "VoodooI2CControllerConstants.hpp"
 
 #define super VoodooI2CController
 OSDefineMetaClassAndStructors(VoodooI2CACPIController, VoodooI2CController);
 
+#define LPSS_PRIVATE_CLOCK_GATING       0x800
+
+void VoodooI2CACPIController::setClkGating(VoodooI2CState enabled) {
+    if (legacy_intel_lpss)
```
kprinssu: Nitpick: We can add an early return instead here?

```c++
if (!legacy_intel_lpss)
  return
  
writeRegister(enabled, LPSS_PRIVATE_CLOCK_GATING);
}
```

VoodooI2C/VoodooI2C/VoodooI2CController/VoodooI2CACPIController.cpp
```diff
@@ -52,6 +62,7 @@ bool VoodooI2CACPIController::start(IOService* provider) {
         return false;
     }
 
+    legacy_intel_lpss = getProperty(kI2CPropIntelLpssKey) != nullptr;
```
kprinssu, line 67: Question: Shouldn't we move this above the `setClkGating` calls?

averycblack: Just to clarify, you mean moving this down to line 80?
The answer is that I probably should. Fairly certain this line executes first already though

kprinssu: My bad, you are absolutely right. Please ignore this comment.

kprinssu:
Another big thanks for the clean up @averycblack

